PR #11, opened 2019-04-18
Add GitHub links to each web package

barryt2:
Add GitHub links to each web packages and remove "NPM Package" term from the title of the package's readme.

### DIFF
--- a/Website/Packages/datax-common/README.md
+++ b/Website/Packages/datax-common/README.md
@@ -1,4 +1,4 @@
-# NPM Package: datax-common
+# datax-common
 
 ### Quick start to developing this package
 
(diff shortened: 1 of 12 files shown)


dineshc-msft:
I see version bump from 1.1.0 to 1.1.1. Is that needed?

Website/Packages/datax-common/README.md
```diff
@@ -1,4 +1,4 @@
-# NPM Package: datax-common
```
carlbrochu: The read me is published to NPM, can you add a section about description of what that the package is used, i.e. include the Data Accelerator top description for each one.

barryt2:
> I see version bump from 1.1.0 to 1.1.1. Is that needed?

Yes because we have to publish our packages in order for the read me updates to show up in NPM. And in order to publish, versions need to be unique (i.e. you cannot publish over existing versions).

barryt2:
#15

Issue #22, opened 2019-04-18
Replace calls to eval in Web Node project


carlbrochu:
**Describe the bug**
Eval can be replaced by a better call

**To Reproduce**

rohit489:
Fixed with #33

PR #33, opened 2019-04-22
remove eval when fethcing the environment variable

rohit489:
#22


Website/Website/config.js
```diff
@@ -12,12 +12,12 @@ function initializeLocalServices() {
     Object.keys(webComposition.localServices).forEach(key => {
```
barryt2: Before doing Object.keys on webComposition.localServices, can you add a if check to check if webComposition.localServices is defined before using it in Object.keys

i.e.
if (webComposition.localServices) {
    Object.keys(...)
    ...
}

rohit489: Done

Issue #32, opened 2019-04-21
Unable to complete the setup


rajmsft:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version of Data Accelerator**
 - Services:
 - DataProcessing
 - WebSite:

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

rajmsft:
> A clear and concise description of what the bug is.

Logged in using windows command prompt as admin
Logged into azure using az login
Updated the parameters with my tenandid and subscription id
deploy.bat always fails with this message
Note: If any module is installed, you will have to close this prompt and restart deploy.bat as admin again...
Install Module:  azuread
Install Module:  mdbc
The script execution completed after one or more packages have been installed. In order to use the latest packages, please close this prompt, open a new command prompt as admin and run deploy.bat again
Error occurred at deployresources.ps1

I tried running it again but same problem

carlbrochu:
Thanks Raj for the feedback!  We have tweaked the experience to work through these types of issues more easily by allowing newer versions of the packages to already be installed, as well as an option to skip the installation of dependency module entirely.  To ensure the scripts can update the components, also make sure to close any potential app using the Azure AAD and mdbc components (i.e. a clean reboot should help).  Thanks and looking forward to having you try the template again.

PR #34, opened 2019-04-22
expose a new param for installing modules and use the minimumVersion for checking

kjcho-msft:
expose a new param for installing modules and use the minimumVersion for checking


kjcho-msft:
#32

PR #35, opened 2019-04-23
bump version of jackson dependency

bryanchen-d:


carlbrochu:
Seems good.  Version changes: 
https://github.com/FasterXML/jackson-core/blob/master/release-notes/VERSION-2.x

PR #43, opened 2019-05-01
add support to simulate data into kafka

rohit489:


Services/DataX.SimulatedData/DataX.SimulatedData.DataGenService/Model/KafkaConnection.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DataX.SimulatedData.DataGenService.Model
+{
+    public class KafkaConnection
+    {
+        public string Broker { get; set; }
```
vijayupadya: Lets rename this to BootstrapServers to make it more accurate

rohit489: Done

PR #53, opened 2019-05-10
DataX.Flow: fix the bugs in the azure function config generation

kjcho-msft:
Changes: 
1. When a code is empty or null, we shouldn't save a secret.
2. Use Id instead of Name which is always empty.

### DIFF
--- a/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveAzureFunction.cs
+++ b/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveAzureFunction.cs
@@ -46,7 +46,7 @@ namespace DataX.Config.ConfigGeneration.Processor
                 foreach (var func in functions.Where(f => f.Type == FunctionTypeName))
                 {
                     var code = func.Properties?.Code;
-                    if (code != null && !KeyVaultUri.IsSecretUri(code))
+                    if (!string.IsNullOrEmpty(code) && !KeyVaultUri.IsSecretUri(code))
                     {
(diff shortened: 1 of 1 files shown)


Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveAzureFunction.cs
```diff
@@ -84,15 +84,14 @@ private AzureFunctionSpec[] GetAzureFuncDefinitions(IList<FlowGuiFunction> funct
             {
                 var properties = f.Properties;
                 Ensure.NotNull(properties, $"guiConfig.process.functions['{f.Id}'].properties");
-                Ensure.NotNull(properties.Name, $"guiConfig.process.functions['{f.Id}'].properties.name");
                 Ensure.NotNull(properties.ServiceEndpoint, $"guiConfig.process.functions['{f.Id}'].properties.serviceEndpoint");
                 Ensure.NotNull(properties.Api, $"guiConfig.process.functions['{f.Id}'].properties.api");
                 Ensure.NotNull(properties.Code, $"guiConfig.process.functions['{f.Id}'].properties.code");
```
vijayupadya: Remove the properties.Code check from here as well since we want to support no key scenario

kjcho-msft: This validation is still valid since properties.Code shouldn't be null. For no key, Code is just empty.

vijayupadya:
Sounds good

PR #56, opened 2019-05-15
rev maven package version

rohit489:


rohit489:
> Are you changing the job definition too?

Created an update

PR #57, opened 2019-05-17
Add kafka input support for data processing

vijayupadya:
- Enable support to take Kafka enabled EventHub and native Kafka as input
- Refactor input object model to make it more extensible
- Add Kafka producer utility app


carlbrochu:
Seems this also breaks the build.  Is that because the binaries are not published?

DataProcessing/datax-host/src/main/scala/datax/input/KafkaStreamingFactory.scala
```diff
@@ -0,0 +1,94 @@
...
+    // EventHub for Kafka needs few additional params
+    if(kafkaInput.isEventHubKafka) {
+      kafkaParams += (
+        "security.protocol" -> kafkaInput.securityProtocol,
+        "sasl.mechanism" -> kafkaInput.saslMechanism,
+        "sasl.jaas.config" -> kafkaInput.saslConfig
+      )
+    }
+
+    //TODO: enable external store based checkpoint
```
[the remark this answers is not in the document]
vijayupadya: Default checkpoint is enabled in Kafka. This comment is about enabling it by writing it to external store for more resiliency.

DataProcessing/datax-keyvault/pom.xml
```diff
@@ -202,7 +202,7 @@ SOFTWARE
               <outputDirectory>${project.build.directory}/mvn-repo</outputDirectory>
               <branch>refs/heads/mvn-repo</branch>
               <includes>
-                <include>**/datax-keyvault_2.11-1.0.0-SNAPSHOT.jar</include>
+                <include>**/*</include>
```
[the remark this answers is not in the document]
vijayupadya: No. This is merge of Rohit's change that is already in.

DataProcessing/datax-core/pom.xml
```diff
@@ -216,7 +216,7 @@ SOFTWARE
               <outputDirectory>${project.build.directory}/mvn-repo</outputDirectory>
               <branch>refs/heads/mvn-repo</branch>
               <includes>
-                <include>**/datax-core_2.11-1.0.0-SNAPSHOT.jar</include>
```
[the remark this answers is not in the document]
vijayupadya: Merged, there was a signature change in AzureFunction that I had to integrate with after that change was merged.

.gitignore
```diff
@@ -6,4 +6,3 @@
 /v15
 npm-debug.log
 /DataProcessing/DataX.Utilities/DataX.Utility.CodeSign/obj
-/Website/Website/deploy/DataX.Utilities/DataX.Utility.CodeSign/obj
```
bryanchen-d: we should add `*/obj` to the ignore list

DataProcessing/datax-host/src/main/scala/datax/input/KafkaInputSetting.scala
```diff
@@ -0,0 +1,142 @@
...
+    bootstrapServers
+  }
+
+  private def getEventHubEndpoint(connStr:String):String ={
+
+    val regex = """sb?://([\w\d\.]+).*""".r
+
+    regex.findFirstMatchIn(connStr) match {
+      case Some(partition) => partition.group(1)
+      case None => null
```
bryanchen-d: shouldn't we throw if the regex doesn't match and we have to return null?

DataProcessing/datax-host/src/main/scala/datax/input/KafkaStreamingFactory.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    kafkaParams
+  }
+
+  def getStream(streamingContext: StreamingContext,
+                inputConf:InputConf,
```
bryanchen-d: instead of passing in a `InputConf`, would be better to pass in the `SettingDictionary`.

bryanchen-d: and the factory parse the flat settings into its known structure of configuration.

DataProcessing/datax-host/src/main/scala/datax/processor/CommonProcessorFactory.scala
```diff
@@ -563,7 +565,40 @@ object CommonProcessorFactory {
...
-      } // end of processPaths
+      }, // end of processPaths
+      /*
+      process a batch of ConsumerRecords from kafka
+     */
+      processConsumerRecord = (rdd: RDD[ConsumerRecord[String,String]], batchTime: Timestamp, batchInterval: Duration, outputPartitionTime: Timestamp) =>{
+        processDataset(rdd
+          .map(d=>{
+            val value = d.value()
+            if(value==null) throw new EngineException(s"null bytes from ConsumerRecord")
```
bryanchen-d: should we just eat the null value and send out a warning instead?

DataProcessing/datax-host/src/main/scala/datax/processor/CommonProcessorFactory.scala
```diff
@@ -563,7 +565,40 @@ object CommonProcessorFactory {
...
+                "timestamp"->d.timestamp().toString,
+                "timestampType"->d.timestampType().toString,
+                "topic"->d.topic()
+              ) ++ key,
+              FileInternal())
+          })
+          .toDF(
+            ColumnName.RawObjectColumn,
+            "Properties",
+            "SystemProperties",
```
bryanchen-d: Since these two constants are used in multiple places, let's add them to the constants list.

DataProcessing/datax-host/src/main/scala/datax/input/KafkaInputSetting.scala
vijayupadya: fixed

DataProcessing/datax-host/src/main/scala/datax/input/KafkaStreamingFactory.scala
vijayupadya: Strongly typed config is a better experience. This way, the parsing of settings dictionary is handled in the corresponding InputSettings class.

DataProcessing/datax-host/src/main/scala/datax/processor/CommonProcessorFactory.scala
vijayupadya: We currently throw if input data is null for eventhub , this change is following the same behavior for kafka as well.

DataProcessing/datax-host/src/main/scala/datax/processor/CommonProcessorFactory.scala
vijayupadya: Updated

PR #66, opened 2019-05-21
Tylake/commonstartup

tylorhl:
### `@`Reviewers
Please review the `ATTENTION` near the bottom of this description for feedback on a behavioral change

## Description Start
This PR looks bigger than it actually is. I refactored `DataXSettings` from `DataX.ServiceHost` to `DataX.Contract` where it makes more sense to live for the future. As such, namespace references changed in a few files.

Here's a breakdown of changes:

### Files of interest

1. `IDataXServiceStartup.cs`
2. `DataXServiceStartup.cs`
3. `DataXApplicationBuilderExtensions.cs`
4. `DataXServiceCollectionExtensions.cs`
5. `DataXWebHostBuilderExtensions.cs`

#### Startup
- **Startup class behavior changes**
  - **`Business logic not changed`**
  - New startup interface: `IDataXServiceStartup` includes [`IStartupFilter`](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.2#extend-startup-with-startup-filters)
  - Utilizing `IStartupFilter` gives DataX the flexibility of being used standalone, or as an addition in other applications when used as a library

- `DataXServiceStartup` provides a common base implementation for other startups to extend without needing to set up common configurations while optionally overriding the configuration call if there are enough differences.
- `DataXServiceStartup` as a common base makes the new startup extensions to be targeted at DataX configuration via type parameter constraints.
- `Startup.cs` files in `Flow` converted to `Flow[FlowServiceName]Startup.cs` for distinguishable use in `.AddDataXStartup<TStartup>()`

#### New Extensions
- New extensions added for use in general configuration
- New extensions added for DataX **Default configurations** based around current setup
  - This is primarily to support running DataX services standalone and as is without additional setup or configuration from others, but is usable as a base configuration for others to utilize
- `DataXServiceCollectionExtensions`
  - Extensions usable on `IServiceCollection`
- `DataXWebHostBuilderExtensions`
  - Extensions usable on `IWebHostBuilder` for ease of DataX configuration without requiring explicit setup in `IServiceCollection` by the consumer
- `DataXApplicationBuilderExtensions`
  - Extensions usable on `IApplicationBuilder` for common configuration

#### `ATTENTION`: Configured Flow
- Changes to the other Flow services were made for common startup, inclusion of `DataXSettings`, and Auth setup
  - **`SLIGHT BEHAVIOR CHANGE: REQUIRES REVIEW`**
    - With the addition of DataX auth to the other Flow services, auth behavior has changed slightly and I need to know if this is okay/desired.
    - Old behavior: Auth required in local scenario
    - New behavior: Auth disabled in local scenario. This behavior matches `Flow.ManagementService`'s behavior.
    - `Flow.ManagementService` is the only service that utilizes the `RolesCheck.Ensure*(isLocal)`. Was the original intention that the other services would also eventually account for local?


Services/DataX.Flow/Flow.InteractiveQueryService/FlowInteractiveQueryServiceStartup.cs
```diff
@@ -6,12 +6,15 @@ namespace Flow.InteractiveQueryService
 {
     using Microsoft.AspNetCore.Hosting;
     using DataX.ServiceHost.AspNetCore;
+    using DataX.ServiceHost.AspNetCore.Startup;
+    using DataX.Contract.Settings;
 
-    public class Startup : StartUpBase
+    public class FlowInteractiveQueryServiceStartup : DataXServiceStartup
```
s-tuli: Can you add comment headers for the new class? ... and in general for all new methods that you have authored in this PR. Thanks!

tylorhl:
> Re: SLIGHT BEHAVIOR CHANGE: REQUIRES REVIEW
> Here are my thoughts:
> Currently we don't support the features from InteractiveQueryService, LiveDataService and SchemaGeneratorService in Local OneBox scenario.
> That said, at some point in the future, we would likely support those features and apis as well especially the InteractiveQueryExperience.
> In general, for local onebox scenario, there is no auth needed. If you look at the tutorial: https://github.com/Microsoft/data-accelerator/wiki/Local-mode-with-Docker#deployment-with-basic-auth, this gives user the option to use basic auth if they so desire for the local onebox scneario.

To be crystal clear, if I understand correctly, your opinion is that this change is not considered harmful and is okay to keep?

Services/DataX.Flow/Flow.ManagementService/FlowManagementServiceStartup.cs
```diff
@@ -0,0 +1,138 @@
...
+                InitialConfiguration.Set(DataX.Config.ConfigDataModel.Constants.ConfigSettingName_ServiceKeyVaultName, Settings.ServiceKeyVaultName);
+            }
+            else
+            {
+                // Local settings
+                var metricsHttpEndpoint = Settings.MetricsHttpEndpoint.TrimEnd('/') + _MetricsHttpEndpointRelativeUri;
+                InitialConfiguration.Set(DataX.Config.ConfigDataModel.Constants.ConfigSettingName_LocalRoot, Settings.LocalRoot);
+                InitialConfiguration.Set(DataX.Config.Local.LocalSparkClient.ConfigSettingName_SparkHomeFolder, Settings.SparkHome);
+                InitialConfiguration.Set(DataX.Config.ConfigDataModel.Constants.ConfigSettingName_ClusterName, "localCluster");
+                InitialConfiguration.Set(DataX.Config.ConfigDataModel.Constants.ConfigSettingName_ServiceKeyVaultName, "local");
```
carlbrochu: Do these need to be fully qualified?  I'd rather avoid having DataX.Config everywhere if possible

tylorhl: Nope. This was a direct copy-paste from the original code before changes occurred. I'll add a using to make it less verbose.

tylorhl:
Changes verified working in ServiceFabric cloud deployment.

Services/DataX.ServiceHost/DataX.ServiceHost.AspNetCore/Extensions/DataXWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,123 @@
...
+    using Microsoft.AspNetCore.Authentication.JwtBearer;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.Extensions.Configuration;
+    using Microsoft.Extensions.DependencyInjection;
+    using System.IO;
+
+    /// <summary>
+    /// Extensions for IWebHostBuilder
+    /// </summary>
+    public static class DataXWebHostBuilderExtensions
```
vijayupadya: Nice!

Services/DataX.ServiceHost/DataX.ServiceHost.AspNetCore/Extensions/DataXServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            startup = startup ?? new TStartup();
+
+            startup.ConfigureServices(services);
+
+            // Add the FlowManagement Startup to be called _along with_ using Configure in the WebHostBuilder
```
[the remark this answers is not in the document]
tylorhl: Good catch

Services/DataX.ServiceHost/DataX.ServiceHost.AspNetCore/Extensions/DataXServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            return services;
+        }
+
+        /// <summary>
+        /// Removes DataX controllers based on a provided DataX startup type
+        /// </summary>
+        /// <typeparam name="TStartup">Concrete startup type for assembly reference</typeparam>
+        /// <param name="manager">The application part manager to remove DataX controllers from</param>
+        /// <returns>The modified application part manager</returns>
+        public static ApplicationPartManager RemoveDataXController<TStartup>(this ApplicationPartManager manager)
```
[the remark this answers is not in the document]
tylorhl: Good question. This is for flexible usage by a consumer.

Let's say I need to wrap my instance of DataX with some custom framework (indeed, this is the case) that has custom ILoggers, auth, etc. I have a couple of options for wrapping:
1. Create multiple projects referencing each of the DataX services with the same scaffolding
2. Reference all of the DataX services in a single project and have the runtime select its role via deployment configuration

In case 2, all DataX controllers for all services would be pulled in by MVC. Controllers can be removed via MVC's `ApplicationPartsManager`. This operation can be done manually by the consumer referencing the assemblies by hand, but I figured a convenience method would be better to have.

As for usage, it's not used within DataX itself.

@vijayupadya Does this help clarify? What are your thoughts?

Services/DataX.Flow/Flow.InteractiveQueryService/Program.cs
```diff
@@ -19,24 +22,42 @@ private static void Main()
         {
             try
             {
```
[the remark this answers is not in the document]
tylorhl: @vijayupadya  Possibly.
Perhaps a static `DataXWebHostRunner` class with overloaded `RunAsync(IWebHostBuilder hostBuilder, State[ful|less]Service service)` that run the same block of code here?

Would this change be appropriate for this PR?

Services/DataX.Flow/Flow.InteractiveQueryService/PackageRoot/Config/Settings.xml
```diff
@@ -11,6 +11,7 @@ Licensed under the MIT License
     <Parameter Name="cosmosDBConfigDatabaseName" Value="" MustOverride="true" />
     <Parameter Name="cosmosDBConfigCollectionName" Value="" MustOverride="true" />
     <Parameter Name="AppInsightsIntrumentationKey" Value="" MustOverride="true" />
+    <Parameter Name="EnableOneBox" Value="" />
```
s-tuli: Update this to:
<Parameter Name="EnableOneBox" Value="" MustOverride="true"/>

tylorhl: I assume you were calling out the missing override? I fixed it. If this is not what you were calling out, please let me know.

s-tuli:
> > Re: SLIGHT BEHAVIOR CHANGE: REQUIRES REVIEW
> > Here are my thoughts:
> > Currently we don't support the features from InteractiveQueryService, LiveDataService and SchemaGeneratorService in Local OneBox scenario.
> > That said, at some point in the future, we would likely support those features and apis as well especially the InteractiveQueryExperience.
> > In general, for local onebox scenario, there is no auth needed. If you look at the tutorial: https://github.com/Microsoft/data-accelerator/wiki/Local-mode-with-Docker#deployment-with-basic-auth, this gives user the option to use basic auth if they so desire for the local onebox scneario.
> 
> To be crystal clear, if I understand correctly, your opinion is that this change is not considered harmful and is okay to keep?



> > Re: SLIGHT BEHAVIOR CHANGE: REQUIRES REVIEW
> > Here are my thoughts:
> > Currently we don't support the features from InteractiveQueryService, LiveDataService and SchemaGeneratorService in Local OneBox scenario.
> > That said, at some point in the future, we would likely support those features and apis as well especially the InteractiveQueryExperience.
> > In general, for local onebox scenario, there is no auth needed. If you look at the tutorial: https://github.com/Microsoft/data-accelerator/wiki/Local-mode-with-Docker#deployment-with-basic-auth, this gives user the option to use basic auth if they so desire for the local onebox scneario.
> 
> To be crystal clear, if I understand correctly, your opinion is that this change is not considered harmful and is okay to keep?

That is correct

Services/DataX.ServiceHost/DataX.ServiceHost.AspNetCore/Extensions/DataXServiceCollectionExtensions.cs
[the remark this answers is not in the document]
vijayupadya: Yes, makes sense.

Services/DataX.Flow/Flow.InteractiveQueryService/Program.cs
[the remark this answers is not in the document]
vijayupadya: Since its a new work, I am ok with deferring it.

PR #68, opened 2019-05-21
add rel="noopener noreferrer" to links

rohit489:
add rel="noopener noreferrer" to links to prevent Reverse Tabnabbing


Website/Website/package.json
```diff
@@ -64,10 +64,10 @@
     "file-loader": "2.0.0",
     "mini-css-extract-plugin": "0.4.4",
     "datax-common": "1.1.1",
-    "datax-home": "1.1.1",
-    "datax-jobs": "1.1.1",
+    "datax-home": "1.2.2-SNAPSHOT-5-21",
+    "datax-jobs": "1.2.2-SNAPSHOT-5-21",
     "datax-metrics": "1.1.1",
```
carlbrochu: do we rev the other components too?

rohit489: updated version of other packages

Website/Packages/datax-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "datax-common",
-  "version": "1.1.1",
+  "version": "1.2.2-SNAPSHOT-5-21",
```
bryanchen-d: shouldn't we remove the SNAPSHOT from the version string?

bryanchen-d: here and the others below

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/input/inputSettingsContent.jsx
```diff
@@ -415,7 +415,7 @@ export default class InputSettingsContent extends React.Component {
                     <Label className="ms-font-m" style={inlineBlockStyle}>
                         Describe Schema in JSON Format
                     </Label>
-                    <a style={linkStyle} href={inputSchemaExampleWiki} target="_blank">
+                    <a style={linkStyle} href={inputSchemaExampleWiki} target="_blank" rel="noopener noreferrer">
```
bryanchen-d: can you check if the `Link` component from office ui fabric has this props by default or not? if yes, I'd suggest we move to use that component. 
If no, first let's search to see if we have used that in our project and also consider we make a similar common `Link` component to contain the shared code.

Website/Packages/datax-common/package.json
rohit489: in master we plan to add snapshot. the version in stable branch wont have snapshot

barryt2: +1, why do we have a snapshot in version name?

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/input/inputSettingsContent.jsx
rohit489: office ui fabric support rel but they dont add rel="noopener noreferrer" by default when target is blank

PR #72, opened 2019-05-31
Adding support for an optional secret for supporting the scenario where the backend services are hosted in an aks cluster

s-tuli:
Adding support for an optional secret for supporting the scenario where the backend services are hosted in an aks cluster. The secret is a key value pair for each servicename and the ip address where the service can be listened at.

### DIFF
--- a/Website/Website/README.md
+++ b/Website/Website/README.md
@@ -62,7 +62,7 @@ faster performance for production consumption.
 npm run build
 ```
 
-### Deploy production version of website with zip-push otpion on Azure Web App
+### Deploy production version of website with zip-push option on Azure Web App
 * Before deploying
 download and install AzureCLI, login first to get access for deployment:
 ```
@@ -84,4 +84,11 @@ set DATAXDEV_GENERATOR_LOCAL_SERVICE=<local URL of Generator service>
 set DATAXDEV_INTERACTIVE_QUERY_LOCAL_SERVICE=<local URL of Interactive Query service>
 set DATAXDEV_SCHEMA_INFERENCE_LOCAL_SERVICE=<local URL of Schema Inference service>
 set DATAXDEV_LIVE_DATA_LOCAL_SERVICE=<local URL of Live Data service>
+```
+
+### (Optional) How to target services hosted on AKS cluster
+If you are hosting DataX services on an AKS cluster, you can configure the website to target the services hosted on an AKS cluster by defining the following secret in the services keyvault. In the secret value: You only need to define the ones you are hosting on AKS cluster.
+The <External IP> should be added for each of the services where they can be listened to on the AKS cluster as specified in the example below:
+```
+{"Flow.InteractiveQueryService":"http://<External IP for Interactive Query Service>:5000","Flow.SchemaInferenceService":"http://<External IP for Schema Inference Service>:5000","Flow.ManagementService":"http://<External IP for Flow Management Service>:5000","Flow.LiveDataService":"http://<External IP for live Data Service>:5000"}
 ```
--- a/Website/Website/README.md
+++ b/Website/Website/README.md
@@ -87,7 +87,9 @@ set DATAXDEV_LIVE_DATA_LOCAL_SERVICE=<local URL of Live Data service>
 ```
 
 ### (Optional) How to target services hosted on AKS cluster
-If you are hosting DataX services on an AKS cluster, you can configure the website to target the services hosted on an AKS cluster by defining the following secret in the services keyvault. In the secret value: You only need to define the ones you are hosting on AKS cluster.
+If you are hosting DataX services on an AKS cluster, you can configure the website to target the services hosted on an AKS cluster by defining the following secret in the kvServices keyvault. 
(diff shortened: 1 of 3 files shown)


Website/Website/auth.js
```diff
@@ -89,6 +89,10 @@ function initialize(host) {
     const authContext = new AuthenticationContext(`https://login.microsoftonline.com/${tenantName}`);
     const serviceClusterUrl = env.serviceClusterUrl;
 
+    // Updates will be made once the ARM support is added.
+    // This expects a secret with a value that is a json object comprising of all the backend service names and the external IP addresses as exposed by AKS where each of them can be listened to.
+    const kubernetesServices = JSON.parse(env.kubernetesServices);
```
kjcho-msft: Can you show an example of the value of env.kubernetesServices?

kjcho-msft: what if env.kubernetesServices is null?

Website/Website/securedSettings.js
```diff
@@ -51,6 +51,10 @@ module.exports = async function(host) {
                     env.mongoDbUrl = await getSecretOrThrow('mongoDbUrl');
                     env.mongoSharedDbUrl = await getSecretOrThrow('mongoSharedDbUrl').catch(err => env.mongoDbUrl);
                 },
+                ,
+                async function() {
+                    env.kubernetesServices = await getSecret('kubernetesServices');
```
kjcho-msft: by looking at the code here, would be better to have this line here
JSON.parse(env.kubernetesServices);

Website/Website/auth.js
s-tuli: Updated the code to do the json parsing of the secret in securedSettings.js

Website/Website/securedSettings.js
s-tuli: updated.

Website/Website/auth.js
```diff
@@ -89,6 +89,10 @@ function initialize(host) {
     const authContext = new AuthenticationContext(`https://login.microsoftonline.com/${tenantName}`);
     const serviceClusterUrl = env.serviceClusterUrl;
 
+    // Updates will be made once the ARM support is added.
+    // This expects a secret with a value that is a json object comprising of all the backend service names and the external IP addresses as exposed by AKS where each of them can be listened to.
+    const kubernetesServices = env.kubernetesServices ? env.kubernetesServices : null;
```
kjcho-msft: can this be just like this?
const kubernetesServices = env.kubernetesServices

Website/Website/README.md
```diff
@@ -85,3 +85,10 @@ set DATAXDEV_INTERACTIVE_QUERY_LOCAL_SERVICE=<local URL of Interactive Query ser
 set DATAXDEV_SCHEMA_INFERENCE_LOCAL_SERVICE=<local URL of Schema Inference service>
 set DATAXDEV_LIVE_DATA_LOCAL_SERVICE=<local URL of Live Data service>
 ```
+
+### (Optional) How to target services hosted on AKS cluster
+If you are hosting DataX services on an AKS cluster, you can configure the website to target the services hosted on an AKS cluster by defining the following secret in the services keyvault. In the secret value: You only need to define the ones you are hosting on AKS cluster.
+The <External IP> should be added for each of the services where they can be listened to on the AKS cluster as specified in the example below:
+```
+{"Flow.InteractiveQueryService":"http://<External IP for Interactive Query Service>:5000","Flow.SchemaInferenceService":"http://<External IP for Schema Inference Service>:5000","Flow.ManagementService":"http://<External IP for Flow Management Service>:5000","Flow.LiveDataService":"http://<External IP for live Data Service>:5000"}
```
kjcho-msft: let's add the secret (kubernetesServices) and the keyvault (this starts with kvServices) name here.

Website/Website/auth.js
barryt2: why do you even need to define a separate constant variable kubernetesServices? You know you can access env.kubernetesServices later in the code, right? I'd suggest removing this unnecessary variable.

Website/Website/auth.js
```diff
@@ -207,6 +212,8 @@ function initialize(host) {
         let url;
         if (env.localServices[query.service]) {
             url = `${env.localServices[query.service]}/api/${query.api}`;
+        } else if (kubernetesServices != null && `${kubernetesServices[query.service]}`) {
```
barryt2: access kubernetesServices as env.kubernetesServices. Also, in javascript, its common to use if (variable) to check for non-null and non-undefined. So in this example, you can simply the statement to be:
"} else if (env.kubernetesServices && env.kubernetesServices[query.services]) {"

Notice in my example above, I removed the `${env.kubernetesServices[query.service]}}`, you don't need it.

Website/Website/auth.js
s-tuli: Removed.

Website/Website/README.md
s-tuli: Readme has been updated

Website/Website/auth.js
s-tuli: updated

PR #78, opened 2019-06-12
Refactor OutputManager to configure outputing non-json data easily

vijayupadya:
* Refactor OutputManager to configure outputing non-json data easily
* Add SqlServer output


DataProcessing/datax-host/src/main/scala/datax/sink/OutputManager.scala
```diff
@@ -95,60 +95,67 @@ object OutputManager {
...
+    val logger = LogManager.getLogger(s"EventsSinker-${sinkerName}")
+    logger.warn(s"Dataframe sinker ${sinkerName} started to sink ${count} events")
+
+
+    // Get the output operators that sink JSON and non-JSON data and call the output sinkers in parallel
+    val jsonOutputOperators = outputOperators.filter(_._3)
+    val nonJsonOutputOperators = outputOperators.filter(!_._3)
+
+    if (count > 0) {
+      Seq("json", "nonjson").par.map {
```
bryanchen-d: why don't we just simply create two `Future`s?

PR #79, opened 2019-06-13
Rev spark version to 2.4 and add suport for secret scope

rohit489:


DataProcessing/datax-host/src/main/scala/datax/securedsetting/KeyVaultClient.scala
```diff
@@ -33,19 +33,25 @@ object KeyVaultClient {
...
         cache.synchronized{
           cache.get(secretId) match {
             case Some(value) => Some(value)
             case None =>
-              val secret = kvc.synchronized{
-                kvc.getSecret(s"https://$vaultName.vault.azure.net",secretName)
-              }
+              var value = ""
+			  if(secretType == "secretscope"){
+				value = dbutils.secrets.get(scope = vaultName, key = secretName)
```
vijayupadya: Recommend synchronizing the calls to dbutils to prevent throttling on secret access, similar to how its done for keyvault below

rohit489: Added synchronized to dbutils

DataProcessing/datax-core/pom.xml
```diff
@@ -51,14 +51,14 @@ SOFTWARE
 
   <groupId>com.microsoft.datax</groupId>
   <artifactId>datax-core_2.3_2.11</artifactId>
-  <version>1.1.0-SNAPSHOT</version>
+  <version>1.2.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
rohit489: Yes it is compatible with HDInsight

Issue #81, opened 2019-06-16
Deploy the iotsample job with running failure


DONGZHUPENG:
After setup Data Accelerator, simply deploy the IoT Sample flow with a new name. 
and several seconds (120s?) later getting below error at jobs tab:

19/06/16 13:56:07 INFO ShutdownHookManager: Shutdown hook called
19/06/16 13:56:07 INFO ShutdownHookManager: Deleting directory /tmp/spark-8dd197f7-3687-4445-93d7-ddb1a1d357ef
19/06/16 13:56:07 INFO MetricsSystemImpl: Stopping azure-file-system metrics system...
19/06/16 13:56:07 INFO MetricsSinkAdapter: azurefs2 thread interrupted.
19/06/16 13:56:07 INFO MetricsSystemImpl: azure-file-system metrics system stopped.
19/06/16 13:56:07 INFO MetricsSystemImpl: azure-file-system metrics system shutdown complete.
stderr:
YARN Diagnostics:
java.lang.Exception: No YARN application is found with tag livy-batch-8-yikjsvds in 120 seconds. Please check your cluster status, it is may be very busy.
org.apache.livy.utils.SparkYarnApp.org$apache$livy$utils$SparkYarnApp$$getAppIdFromTag(SparkYarnApp.scala:182) org.apache.livy.utils.SparkYarnApp$$anonfun$1$$anonfun$4.apply(SparkYarnApp.scala:239) org.apache.livy.utils.SparkYarnApp$$anonfun$1$$anonfun$4.apply(SparkYarnApp.scala:236) scala.Option.getOrElse(Option.scala:120) org.apache.livy.utils.SparkYarnApp$$anonfun$1.apply$mcV$sp(SparkYarnApp.scala:236) org.apache.livy.Utils$$anon$1.run(Utils.scala:97)

carlbrochu:
Thanks you for trying Data Accelerator and reporting this issue.  Apologies for running into this issue.  This error usually mean there mismatch in the environment between the configuration and the deployed resources.  We are reviewing the data and will get back to you as soon as possible with more details on how to fix this issue.  Stay tuned.

carlbrochu:
Thanks again for reporting this issue, we are making an update to the template supporting code to avoid this issue in the future.  

To fix this issue in your environment, you will need to update the default job settings.  To do so: 
[...]

PR #82, opened 2019-06-17
ARM: remove snapshot from the templates to unblock ARM deployment fro…

kjcho-msft:
…m master branch


carlbrochu:
Thanks Kwangje.  #81

PR #84, opened 2019-06-18
Refactor Query component into its own package

s-tuli:
1. Updating the website code to extract out the Query package named dataX-query. This datax-query package content used to be part of the package datax-pipeline. The reason that this package is being extracted out is such that this package can be used by othercustomers who do not want a dependency on the datax-pipeline package.
2. Also fixing the versions in the servicemanifest.xml to fix the local development scenario for the backend services under DataX.Flow
3. Fixing a warning due to a required DataBricks parameter


Services/DataX.Flow/Flow.SchemaInferenceService/PackageRoot/ServiceManifest.xml
```diff
@@ -4,7 +4,7 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 Licensed under the MIT License
 -->
 <ServiceManifest Name="Flow.SchemaInferenceServicePkg"
-                 Version="1.0.1"
+                 Version="1.0.0"
```
carlbrochu: why going back to .0?

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
```diff
@@ -733,11 +737,11 @@ class FlowDefinitionPanel extends React.Component {
     onResampleInput(kernelId) {
         const version = this.props.kernelVersion + 1;
         this.props.onUpdateKernelVersion(version);
-        this.props.onResampleInput(this.props.flow, kernelId, version);
+        this.props.onResampleInput(this.props.flow.name, this.props.flow.displayName, this.props.flow.owner, this.props.flow.input.properties.inputSchemaFile, this.props.flow.input.properties.normalizationSnippet, this.props.flow.input.properties.inputEventhubConnection, this.props.flow.input.properties.inputSubscriptionId, this.props.flow.input.properties.inputResourceGroup, this.props.flow.input.properties.inputEventhubName, this.props.flow.input.type, this.props.flow.resamplingInputDuration, kernelId, version);
```
carlbrochu: That s a lot of params.  Should we pass in the Flow object instead?

Website/Packages/datax-query/src/modules/query/queryModels.js
```diff
@@ -0,0 +1,26 @@
...
+        return defaultQuery;
+    }
+}
+//User can use the sample or tutorial or intellisense to have starting query. Below allows default to have 5 blank lines.
+export const defaultQuery = `
+
+`;
+
+export const defaultQueryLocal = `--DataXQuery--
+events = SELECT MAX(temperature) as maxTemp
```
carlbrochu: Is this used anywhere?  I thought we had a different default.

Services/DataX.Flow/Flow.SchemaInferenceService/PackageRoot/ServiceManifest.xml
s-tuli: This change actually fixes a bug where the version needs to be consistent across various manifest files. This version exists in various other files as well viz. ApplicationManifest.xml and all these files must have the same version. re: why 1.0.0?, IMO, there is no need to rev up the version. It should not  have been changed to 1.0.1 for partial files. This change fixes the break which blocks the local-cloud deployment of service fabric backend services scenario for developers.

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
s-tuli: This is an intentional change. It used to pass in the flow object perviously since all of this code was in the same package: datax-pipeline package. If we had continued with that design. that would mean that the datax-query package will have dependency on the datax-pipeline package. This PR creates a dependency for the datax-pipeline package on the datax-query package.  What you suggest will 1. create a circular dependency. 2. When another customer uses this package, they will need to have a dependency on the datax-pipeline package which we do not want. In fact, any of these params can be optional. I have currently made rules as optional since rules are very specific to data-accelerator

Website/Packages/datax-query/src/modules/query/queryModels.js
s-tuli: Yes.
Used by the below function:
export function getDefaultQuery(enableLocalOneBox) {... line 6 in queryModels.js
... which is used in the local onebox experience.

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
carlbrochu: Interesting.  So it is because Flow is defined in Pipeline.  Could Flow be moved too at some point into common components?

Website/Packages/datax-query/src/modules/query/queryActions.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ *
+ * REDUX Action Types
+ *
+ */
+
+// Init
+export const QUERY_INIT = 'QUERY_INIT';
+export const QUERY_NEW = 'QUERY_NEW';
```
barryt2: There is currently no "action" functions using QUERY_INIT and QUERY_NEW, you should remove this dead code

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
```diff
@@ -37,8 +32,17 @@ import {
     IconButtonStyles,
     VerticalTabs,
     VerticalTabItem,
-    getApiErrorMessage
+    getApiErrorMessage,
```
barryt2: nit: remove ending comma

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
barryt2: We should not need to move flow into the datax-query package as that data to my understanding is heavily coupled to the implementation details of the datax-pipeline package. For example, you wouldn't want to need to modify the flow object defined in the datax-query package just to introduce new state/behavior of the datax-pipeline package. Also, for the other customer that you know I'm talking about, they don't care about flows at all.

barryt2: I do agree that there are a lot of params, Anu will be addressing this by formalizing a params object to wrap the min requirements to pass in the functions as a single arg instead of N args.

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowHelpers.js
```diff
@@ -4,13 +4,6 @@
...
 import * as Models from './flowModels';
 
-export function isValidNumberAboveZero(value) {
-    const number = Number(value);
-    const isNumber = !isNaN(number);
-    const isValid = isNumber && number > 0 && value[0] !== '0';
-    return isValid;
-}
-
 export function isValidNumberAboveOrEqualZero(value) {
```
barryt2: If we are moving isValidNumberAboveZero to commons, does it also make sense to move isValidNumberAboveOrEqualZero to commons too? They seem similar/related and other future package developers might benefit from this common helper function.

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowReducer.js
```diff
@@ -189,7 +190,7 @@ export default (state = INITIAL_FLOW_STATE, action) => {
         case Actions.FLOW_UPDATE_SELECTED_FUNCTION_INDEX:
             return Object.assign({}, state, { selectedFunctionIndex: action.payload });
 
-        case Actions.FLOW_UPDATE_QUERY:
+        case QueryActions.QUERY_UPDATE_QUERY:
```
barryt2: It looks like this is a bug where you are handling and saving the query state here in the flow reducer from datax-pipeline and the query reducer from datax-query packages. This also applies above (in this file) for QUERY_UPDATE_RESAMPLING_INPUT_DURATION

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
s-tuli: Updated

s-tuli: Updated.

Website/Packages/datax-query/src/modules/query/queryActions.js
s-tuli: Updated

Website/Packages/datax-query/src/modules/query/queryReducer.js
```diff
@@ -0,0 +1,46 @@
...
+import * as Models from './queryModels';
+
+const INITIAL_QUERY_STATE = {    
+    query: Models.defaultQuery,   
+};
+
+export default (state = INITIAL_QUERY_STATE, action) => {
+    switch (action.type) {
+        
+        case Actions.QUERY_INIT:
```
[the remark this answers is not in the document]
s-tuli: Updated

Website/Packages/datax-query/src/modules/query/queryReducer.js
```diff
@@ -0,0 +1,46 @@
...
+            const flow = action.payload;
```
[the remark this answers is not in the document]
s-tuli: Updated

Website/Packages/datax-query/src/modules/query/querySelectors.js
```diff
@@ -0,0 +1,26 @@
...
+
+
+// Validation - Query
+export const validateFlowQuery = createSelector(
+    getFlowQuery,
+    validateQuery
+);
+
+function validateQuery(query) {
+    //removing validation; codegen will add OUTPUTs for alerts. Blank query is valid.
```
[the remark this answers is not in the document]
s-tuli: Updated

Services/DataX.Flow/Flow.InteractiveQueryService/PackageRoot/ServiceManifest.xml
```diff
@@ -4,7 +4,7 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 Licensed under the MIT License
 -->
 <ServiceManifest Name="Flow.InteractiveQueryServicePkg"
-                 Version="1.0.1"
+                 Version="1.0.0"
```
bryanchen-d: curious why are we rolling back the version?

Website/Packages/datax-query/src/modules/query/queryActions.js
```diff
@@ -0,0 +1,112 @@
...
+    });
+};
+
+export const getCodeGenQuery = (flowname, displayName, query, rules={}, outputTemplates) => {
+    return Api.getCodeGenQuery(flowname, displayName, query, rules, outputTemplates).then(query => {
+        return query;
+    });
+};
+
+export const executeQuery = (flowname, displayName, selectedQuery, rules={}, outputTemplates, kernelId) => {
```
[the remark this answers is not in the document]
s-tuli: In any case updated to use an object instead.

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowHelpers.js
s-tuli: Currently only isValidNumberAboveZero is being used by 2 packages. IMO, we can move the functions to common as needed if they are being used by multiple packages. This is not the only one function in the FlowHelpers.js which could potentially be moved. Thanks!

Services/DataX.Flow/Flow.InteractiveQueryService/PackageRoot/ServiceManifest.xml
s-tuli: It was mistakenly checked in. The version was rev'd was only partially for servicemanifest.xml files under dataX.Flow whereas the ApplicationManifest.xml was still having the version 1.0.0. This broke the local debugging scenario for DataX.Flow. Hence fixing the bug by reverting the version in servicemanifest.xml for each of the services since the version needs to be consistent across.

Website/Packages/datax-query/src/modules/query/queryReducer.js
```diff
@@ -0,0 +1,46 @@
...
+            });
+        
+        case Actions.QUERY_UPDATE_QUERY:
+            return Object.assign({}, state, {
+                isDirty: true,
+                query: action.payload
+            });
+
+        case Actions.QUERY_UPDATE_ERROR_MESSAGE:
+            return Object.assign({}, state, { errorMessage: action.message });
```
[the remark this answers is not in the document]
s-tuli: Updated

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
s-tuli: Updated. (I would have expected this to throw a compilation error)

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowReducer.js
s-tuli: updated

Website/Packages/datax-query/src/styles/styles.css
```diff
@@ -0,0 +1,27 @@
+/*
+ *********************************************************************
+ Copyright (c) Microsoft Corporation.  All rights reserved.
+ Licensed under the MIT License
+ *********************************************************************
+*/
+
+/* Query Pane styles */
+.query-pane-button {
```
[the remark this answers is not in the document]
s-tuli: Updated the style.css under datax-pipeline package.

Website/Packages/datax-query/src/styles/styles.css
```diff
@@ -0,0 +1,27 @@
...
+    height: 27px !important;
+    padding-left: 5px !important;
+    padding-right: 5px !important;
+    min-width: 1px !important;
+  }
+
+.ms-IconExample:hover {
```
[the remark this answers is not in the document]
s-tuli: yes. line number: 227 in querySettingsContent.jsx

Website/Packages/datax-query/src/modules/query/queryModels.js
```diff
@@ -0,0 +1,26 @@
+// *********************************************************************
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License
+// *********************************************************************
+
+export function getDefaultQuery(enableLocalOneBox) {
```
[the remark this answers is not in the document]
s-tuli: Updated

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
barryt2: this doesn't throw an error in C# either. Its pretty much the same behavior across several languages such as C++, Java, etc.

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
```diff
@@ -782,18 +788,18 @@ const mapStateToProps = state => ({
...
+    flowValidated: Selectors.validateFlow(state),
+    queryValidated: QuerySelectors.validateQueryTab(state)
 });
 
 // Dispatch Props
 const mapDispatchToProps = dispatch => ({
     // Init Actions
     initFlow: context => dispatch(Actions.initFlow(context)),
-
+    initQuery: query =>dispatch(QueryActions.initQuery(query)),
```
barryt2: nit: put a space between => and dispatch

barryt2: prettifier should've taken care of this, did you not open the datax-pipeline package folder when doing the edits?

Website/Packages/datax-query/src/modules/query/querySelectors.js
```diff
@@ -0,0 +1,26 @@
+// *********************************************************************
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License
+// *********************************************************************
+import { createSelector } from 'reselect';
+
+// Settings - Flow
+export const getQuery = state => state.query;
```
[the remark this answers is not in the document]
s-tuli: Updated

Website/Packages/datax-query/package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "rimraf dist && webpack --mode production --config webpack.config.js --bail --progress --display-modules --display-error-details --colors --display-chunks --display-reasons",
+    "dev": "rimraf dist && webpack --mode development --config webpack.config.js --bail --progress --display-modules --display-error-details --colors --display-chunks --display-reasons",
+    "devpatch": "rimraf dist && webpack --mode development --config webpack.config.js --bail --progress --display-modules --display-error-details --colors --display-chunks --display-reasons && localdevpatch.bat"
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "@babel/core": "7.1.6",
```
kjcho-msft: do we need all these packages?

s-tuli: I believe, yes

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/flowDefinitionPanel.jsx
s-tuli: updated. Prettifier is not working.

PR #85, opened 2019-06-18
Refactored the Dockerfile and finalrun.sh such that CICD can be enabled

s-tuli:
- Refactored the Dockerfile and finalrun.sh such that CICD can be enabled easily by passing in the service name as parameter. 
- Also adding the yaml files for each service which will need the parameters to be passed in prior to deploying the service to the Kubernetes cluster (just as we do for when deploying to the service fabric cluster.)
Testing has been done by using the kubectl command for each of the services.


carlbrochu:
Looks good

PR #90, opened 2019-06-24
Enable batch processing of blob input

vijayupadya:


DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
```diff
@@ -432,6 +432,7 @@ object HadoopClient {
     * @return a list of file paths under the folder
     */
   def listFiles(folder: String): Iterator[String] = {
+    resolveStorageAccountKeyForPath(folder)
```
bryanchen-d: better to resolve the key before calling listFiles, so we can decouple hadoopClient and keyvault

vijayupadya: We already have this pattern of resolving storage in other methods in hadoopclient like writeHdfsFile/createFolder. This is adopting this same pattern for listing.

PR #94, opened 2019-07-01
Fixing and adding a check for Databricks vs HDInsight when saving and resolving the spark job params

s-tuli:
Fixing and adding a check for Databricks vs HDInsight when saving and resolving the spark job params

### DIFF
--- a/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveSparkJobParams.cs
+++ b/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveSparkJobParams.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Composition;
 using System.Text;
 using System.Threading.Tasks;
+using DataX.Config.ConfigDataModel;
 
 namespace DataX.Config.ConfigGeneration.Processor
 {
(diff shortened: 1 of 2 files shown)


Services/DataX.Flow/DataX.Flow.Common/Helper.cs
```diff
@@ -56,7 +56,7 @@ public static string GetSecretFromKeyvaultIfNeeded(string value)
         /// <returns>true if it is a secret, otherwise false</returns>
         public static bool IsKeyVault(string value)
         {
-            return (value.StartsWith(GetKeyValutNamePrefix()) || value.StartsWith(GetSecretScopePrefix()));
+            return value.StartsWith(GetKeyValutNamePrefix()) || value.StartsWith(GetSecretScopePrefix());
```
carlbrochu: why change this one?

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveSparkJobParams.cs
```diff
@@ -26,10 +27,17 @@ public class ResolveSparkJobParams: ProcessorBase
         public const string TokenName_DatabricksToken = "guiSparkDatabricksToken";
         public const string TokenName_SparkJobDatabricksAutoScale = "guiSparkJobDatabricksAutoScale";
 
+        [ImportingConstructor]
+        public ResolveSparkJobParams(ConfigGenConfiguration conf)
```
carlbrochu: does the call to the ctor need to change?

Services/DataX.Flow/DataX.Flow.Common/Helper.cs
s-tuli: I was seeing a formating error, hence fixing it in this checkin

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S500_ResolveSparkJobParams.cs
s-tuli: No caller to the constructor.as such.

PR #98, opened 2019-07-09
Update resource files to have unit tests pass again.

carlbrochu:
Update resources for unit tests to pass again.


Services/DataX.Config/DataX.Config.Local.Test/Resource/flowSaved.json
```diff
@@ -332,7 +335,7 @@
           "properties": {
             "name": "myudaf",
             "path": "wasbs://sdfsdf/sdfsdf",
-            "class": "datax.udaf.myudaf",
+            "class": "datax.udf.myudaf",
```
kjcho-msft: I think this is fine. Another option is that we can replace "class": "datax.udf.myudaf" with "class": "datax.udaf.myudaf" in  flowStarted.json since this is for an udaf section. :)

carlbrochu: good point, i'll leave it as is for now and see if we need to make further changes down the road

PR #102, opened 2019-07-17
bug fixes: spark nuspec, iothub sku, simulator service num events

kjcho-msft:
Update spark nuspec to include azure-sqldb-spark
Change ARM template iothub sku to B1
Change the simulator service num of events to 50

### DIFF
--- a/DeploymentCloud/Deployment.DataX/Samples/samples/iotDevice/iotsample.json
+++ b/DeploymentCloud/Deployment.DataX/Samples/samples/iotDevice/iotsample.json
@@ -4,7 +4,7 @@
 		{
 			"dataTypeName": "DeviceData",
 			"simulationPeriodInMinute": 1,
-			"numEventsPerBatch": 100,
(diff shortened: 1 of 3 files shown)


DeploymentCloud/Deployment.DataX/Samples/samples/iotDevice/iotsample.json
```diff
@@ -4,7 +4,7 @@
 		{
 			"dataTypeName": "DeviceData",
 			"simulationPeriodInMinute": 1,
-			"numEventsPerBatch": 100,
+			"numEventsPerBatch": 50,
```
carlbrochu: how many events is that per month, will that work well with the change above?  Can we document these changes either in this PR or in our tutorials/wiki pages?

kjcho-msft: With this change, about 78,000 events will be sent. It will be good enough. Worked with Rohit for this.

PR #104, opened 2019-07-24
Fix unit tests

rohit489:


Services/DataX.Config/DataX.Config.Local.Test/Resource/flowSaved.json
```diff
@@ -294,6 +294,7 @@
   "gui": {
     "name": "localconfiggentest",
     "displayName": "localconfiggentest",
+    "databricksToken": null,
```
carlbrochu: can we also have a test that doesn't have these defined?

Services/DataX.Config/DataX.Config.Test/FlowOperationTest.cs
```diff
@@ -27,11 +28,13 @@ public class FlowOperationTest
         public static void Initialize(TestContext tc)
         {
             InitialConfiguration.Set(Constants.ConfigSettingName_RuntimeKeyVaultName, "somekeyvault");
+            InitialConfiguration.Set(Constants.ConfigSettingName_SparkType, "hdinsight");
```
carlbrochu: is this optional?

rohit489: made it optional and removed it from hdInsights testcases. Added another end to end test case for databricks and setting sparktype in that test case

Services/DataX.Config/DataX.Config.Local.Test/Resource/flowSaved.json
rohit489: going forward save api will always add this. based on sparktype this will either have a value or be null

Services/DataX.Config/DataX.Config.Local.Test/Resource/flowSaved.json
```diff
@@ -69,150 +69,6 @@
           },
           "dynamicOffsetInMs": 300000
         }
-      },
```
kjcho-msft: Did you fix fix the local E2E test as well?

rohit489: the local E2E test was originally disabled so dint touch that

PR #105, opened 2019-07-30
For databricks access azure storage account using account key from keyvault

rohit489:
For databricks access azure storage account using account key from keyvault.

### DIFF
--- a/DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
+++ b/DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
@@ -127,8 +127,15 @@ object HadoopClient {
       case Some(value)=>
         logger.warn(s"Retrieved key for storage account '$sa' with secretid:'$secretId'")
         setStorageAccountKey(sa, value)
-      case None =>
-        logger.warn(s"Failed to find key for storage account '$sa' with secretid:'$secretId'")
+      case None =>	    
+        val databricksSecretId = s"secretscope://$vaultName/${ProductConstant.ProductRoot}-sa-$sa"
+		KeyVaultClient.getSecret(databricksSecretId) match {
+		  case Some(value)=>
+            logger.warn(s"Retrieved key for storage account '$sa' with secretid:'$databricksSecretId'")
+            setStorageAccountKey(sa, value)
+		  case None =>
+		    logger.warn(s"Failed to find key for storage account '$sa' with secretid:'$secretId' and '$databricksSecretId'")
+		}
     }
   }
 
--- a/DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
+++ b/DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
@@ -122,6 +122,7 @@ object HadoopClient {
     * @param sa name of the storage account
     */
   private def resolveStorageAccount(vaultName: String, sa: String) = {
+    // Fetch secret from keyvault using KeyVaultMsiAuthenticatorClient and if that does not return secret then fetch it using secret scope  
     val secretId = s"keyvault://$vaultName/${ProductConstant.ProductRoot}-sa-$sa"
     KeyVaultClient.getSecret(secretId) match {
       case Some(value)=>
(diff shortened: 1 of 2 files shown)


DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
```diff
@@ -127,8 +127,15 @@ object HadoopClient {
       case Some(value)=>
         logger.warn(s"Retrieved key for storage account '$sa' with secretid:'$secretId'")
         setStorageAccountKey(sa, value)
-      case None =>
-        logger.warn(s"Failed to find key for storage account '$sa' with secretid:'$secretId'")
+      case None =>	    
+        val databricksSecretId = s"secretscope://$vaultName/${ProductConstant.ProductRoot}-sa-$sa"
```
carlbrochu: can you add a comment to detail what and why we are doing this?  Also, is this a possibility to add a unit test here?

rohit489: Added comment. Adding unit tests will be done as a separate work item

DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
```diff
@@ -122,13 +122,21 @@ object HadoopClient {
     * @param sa name of the storage account
     */
   private def resolveStorageAccount(vaultName: String, sa: String) = {
+    // Fetch secret from keyvault using KeyVaultMsiAuthenticatorClient and if that does not return secret then fetch it using secret scope  
     val secretId = s"keyvault://$vaultName/${ProductConstant.ProductRoot}-sa-$sa"
```
kjcho-msft: Who creates this secret ${ProductConstant.ProductRoot}-sa-$sa? Do we have a corresponding service change to create it? 

For now, in DataX.Flow, we currently create this kind of secret for two cases- 1. when the input type is a blob 2. when a blob output is used. Is this good enough?

kjcho-msft: Curious to know. What if there is a key like this for a databricks environment? Will it stiill work?

DataProcessing/datax-host/src/main/scala/datax/host/SparkJarLoader.scala
```diff
@@ -57,7 +59,9 @@ object SparkJarLoader {
 
   def addJar(spark: SparkSession, jarPath: String) = {
     addJarOnDriver(spark, jarPath)
-    spark.sparkContext.addJar(jarPath)
+	val logger = LogManager.getLogger("AddJar to executer")
+    logger.info("fileUrl is " + fileUrl)
+    spark.sparkContext.addJar(fileUrl)
```
kjcho-msft: Does this still work for Spark?

kjcho-msft: Adding a comment?

DataProcessing/datax-host/src/main/scala/datax/host/SparkJarLoader.scala
```diff
@@ -23,6 +23,7 @@ import scala.collection.mutable.HashMap
 
 object SparkJarLoader {
   val currentJars = new HashMap[String, Long]
+  var fileUrl = ""
```
vijayupadya: Make fileUrl as return param of function addJarOnDriver and remove it as global variable.

DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
rohit489: This secret is already being created by flow service. No further change is required in service

rohit489: secret with name "${ProductConstant.ProductRoot}-sa-$sa" is created in keyvault. CP will try to first fetch the secret using KeyVaultMsiAuthenticatorClient which is used in HDInsight environment. For databricks environment, we read the same secret from the keyvault but the way to read a secret from keyvault in databrick is through secret scope. By adding the prefix 'secretscope' instead of 'keyvault', we tell CP to use secret scope through databricks utility to access the secret from keyvault since in databricks we cannot directly access secrets in keyvault

DataProcessing/datax-host/src/main/scala/datax/host/SparkJarLoader.scala
rohit489: verified that it works with HDInsight and databricks. Also added a comment

DataProcessing/datax-host/src/main/scala/datax/host/SparkJarLoader.scala
rohit489: Done

PR #106, opened 2019-07-31
Refactor DataX.Flow to support batch scenarios better

kjcho-msft:
We don't need to generate Transform, Projection and Schema files when there is no job to deploy so the following will be executed after GenerateJobConfig and GenerateJobConfigBatch.
GenerateTransformFile
GenerateProjectionFile
GenerateSchemaFile

### DIFF
--- a/Services/DataX.Config/DataX.Config.DatabricksClient/DatabricksClient.cs
+++ b/Services/DataX.Config/DataX.Config.DatabricksClient/DatabricksClient.cs
@@ -156,6 +156,7 @@ namespace DataX.Config.DatabricksClient
 
         public async Task<SparkJobSyncResult[]> GetJobs()
         {
+            await Task.CompletedTask;
             throw new NotImplementedException();
         }
     }
--- a/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
+++ b/Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
@@ -18,7 +18,7 @@ namespace DataX.Config.ConfigGeneration.Processor
     /// </summary>
     [Shared]
     [Export(typeof(IFlowDeploymentProcessor))]
-    public class GenerateTransformFile : ProcessorBase
+    public class PrepareTransformFile : ProcessorBase
     {
         public const string TokenName_TransformFile = "processTransforms";
         public const string AttachmentName_CodeGenObject = "rulesCode";
@@ -28,7 +28,7 @@ namespace DataX.Config.ConfigGeneration.Processor
         private IRuntimeConfigStorage RuntimeStorage { get; }
 
         [ImportingConstructor]
-        public GenerateTransformFile(IKeyVaultClient keyvaultClient, IRuntimeConfigStorage runtimeStorage, ConfigGenConfiguration conf)
+        public PrepareTransformFile(IKeyVaultClient keyvaultClient, IRuntimeConfigStorage runtimeStorage, ConfigGenConfiguration conf)
         {
             KeyVaultClient = keyvaultClient;
             RuntimeStorage = runtimeStorage;
@@ -51,7 +51,7 @@ namespace DataX.Config.ConfigGeneration.Processor
             }
             string queries = string.Join("\n", guiConfig.Process?.Queries);
 
-            string ruleDefinitions =  RuleDefinitionGenerator.GenerateRuleDefinitions(guiConfig.Rules, config.Name);
+            string ruleDefinitions = RuleDefinitionGenerator.GenerateRuleDefinitions(guiConfig.Rules, config.Name);
             RulesCode rulesCode = CodeGen.GenerateCode(queries, ruleDefinitions, config.Name);
 
             Ensure.NotNull(rulesCode, "rulesCode");
@@ -59,18 +59,17 @@ namespace DataX.Config.ConfigGeneration.Processor
             // Save the rulesCode object for downstream processing
             flowToDeploy.SetAttachment(AttachmentName_CodeGenObject, rulesCode);
 
-            var runtimeConfigBaseFolder = flowToDeploy.GetTokenString(PrepareJobConfigVariables.TokenName_RuntimeConfigFolder);
-            Ensure.NotNull(runtimeConfigBaseFolder, "runtimeConfigBaseFolder");
-
             var runtimeKeyVaultName = flowToDeploy.GetTokenString(PortConfigurationSettings.TokenName_RuntimeKeyVaultName);
             Ensure.NotNull(runtimeKeyVaultName, "runtimeKeyVaultName");
 
-            var filePath = ResourcePathUtil.Combine(runtimeConfigBaseFolder, $"{config.Name}-combined.txt");
-            var transformFilePath = await RuntimeStorage.SaveFile(filePath, rulesCode.Code);
             var secretName = $"{config.Name}-transform";
-            var transformFileSecret = await KeyVaultClient.SaveSecretAsync(runtimeKeyVaultName, secretName, transformFilePath, Configuration.TryGet(Constants.ConfigSettingName_SparkType, out string sparkType) ? sparkType : null);
+            Configuration.TryGet(Constants.ConfigSettingName_SparkType, out string sparkType);
+            var uriPrefix = (sparkType != null && sparkType == Constants.SparkTypeDataBricks) ? Constants.PrefixSecretScope : Constants.PrefixKeyVault;
+
+            var transformFileSecret = $"{uriPrefix}://{runtimeKeyVaultName}/{secretName}";
             flowToDeploy.SetStringToken(TokenName_TransformFile, transformFileSecret);
 
+            await Task.CompletedTask;
             return "done";
         }
     }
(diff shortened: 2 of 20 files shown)


Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            var runtimeKeyVaultName = flowToDeploy.GetTokenString(PortConfigurationSettings.TokenName_RuntimeKeyVaultName);
+            Ensure.NotNull(runtimeKeyVaultName, "runtimeKeyVaultName");
+
+            var secretName = $"{config.Name}-transform";
+            Configuration.TryGet(Constants.ConfigSettingName_SparkType, out string sparkType);
+            var uriPrefix = (sparkType != null && sparkType == Constants.SparkTypeDataBricks) ? Constants.PrefixSecretScope : Constants.PrefixKeyVault;
+
+            var transformFileSecret = $"{uriPrefix}://{runtimeKeyVaultName}/{secretName}";
+            flowToDeploy.SetStringToken(TokenName_TransformFile, transformFileSecret);
```
vijayupadya: Where is this secret saved to keyvault? In general it is not a good practice to have the secret token set without saving the secret. If the secret save fails, we will have a non-existent key set in the token.

Services/DataX.Utilities/DataX.Utilities.Blob/BlobHelper.cs
```diff
@@ -153,9 +153,9 @@ public static async Task<List<string>> GetLastModifiedBlobContentsInBlobPath(str
             CloudBlobClient cloudBlobClient = storageAccount.CreateCloudBlobClient();
             CloudBlobContainer container = cloudBlobClient.GetContainerReference(containerName);
 
-            var filteredBlobs = container.ListBlobsSegmentedAsync(prefix: prefix, useFlatBlobListing: true, blobListingDetails: BlobListingDetails.None, maxResults: null, currentToken: null, options: null, operationContext: null).Result.Results.OfType<CloudBlockBlob>()
+            var filteredBlobs = await Task.Run(() => container.ListBlobsSegmentedAsync(prefix: prefix, useFlatBlobListing: true, blobListingDetails: BlobListingDetails.None, maxResults: null, currentToken: null, options: null, operationContext: null).Result.Results.OfType<CloudBlockBlob>()
```
vijayupadya: Why this change? i.e invoking the listBlobs in a separate task?

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S600_GenerateJobConfigBatch.cs
```diff
@@ -224,7 +224,7 @@ private async Task<JobConfig> ScheduleSingleJob(JobDeploymentSession job, string
             destFolder = GetJobConfigFilePath(isOneTime, dateString, destFolder);
             var jc = new JobConfig
             {
-                Content = await GetBatchConfigContent(job, defaultJobConfig.ToString(), processStartTime, processEndTime).ConfigureAwait(false),
+                Content = GetBatchConfigContent(job, defaultJobConfig.ToString(), processStartTime, processEndTime),
```
vijayupadya: Why async is removed?

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
kjcho-msft: Saving this secret to KV does happen in S610_GenerateTransformFile.cs. I made it on purpose. since if we don't have any job to deploy, we shouldn't save any file and even any secret.

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S600_GenerateJobConfigBatch.cs
kjcho-msft: I realized that it doesn't need to be async call.

Services/DataX.Utilities/DataX.Utilities.Blob/BlobHelper.cs
kjcho-msft: If there are a lot of blobs in the target folder, it could probably affect the performance. Having this would help. Any concern? I agree it can be non-async call though.

Services/DataX.Config/DataX.Config.DatabricksClient/DatabricksClient.cs
```diff
@@ -156,6 +156,7 @@ public async Task<SparkJobSyncResult> StopJob(JToken jobClientData)
 
         public async Task<SparkJobSyncResult[]> GetJobs()
         {
+            await Task.CompletedTask;
             throw new NotImplementedException();
```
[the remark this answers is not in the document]
kjcho-msft: Actually the code in this class was not part of my changes. This was added by Rohit. This is just a fix for bug 956780. @rohit489 can clarify this.

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
```diff
@@ -0,0 +1,76 @@
...
+using DataX.Config.ConfigDataModel;
+
+namespace DataX.Config.ConfigGeneration.Processor
+{
+    /// <summary>
+    /// Produce the transform/query file section
+    /// </summary>
+    [Shared]
+    [Export(typeof(IFlowDeploymentProcessor))]
+    public class PrepareTransformFile : ProcessorBase
```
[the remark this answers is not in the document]
kjcho-msft: Existing config E2E tests are good enough, i mean, checking the job config output is good enough. Will add a comment like this. 
        /// <summary>
        /// Generate and set the info for the transform file which will be needed to generate JobConfig
        /// </summary>

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
kjcho-msft: And the reason why I have to compose this secret url here is that the url is needed to generate the job config which is s600.

kjcho-msft: updated to use the secret url which was saved from the prepare steps.

Services/DataX.Utilities/DataX.Utilities.Blob/BlobHelper.cs
kjcho-msft: update not to use task.run just use await.

Services/DataX.Config/DataX.Config.DatabricksClient/DatabricksClient.cs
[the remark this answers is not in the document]
rohit489: We never call into GetJobs hence this was not implemented. However GetJobs is part of the ISparkJobClient hence it needed to be added to DatabricksClient class. In future if we ever require a GetJobs functionality then this method can be implemented at that time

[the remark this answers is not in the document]
kjcho-msft: Thanks, Rohit!

Services/DataX.Config/DataX.Config/ConfigGeneration/Processor/S450_PrepareTransformFile.cs
```diff
@@ -51,26 +54,25 @@ public override async Task<string> Process(FlowDeploymentSession flowToDeploy)
...
-
             var runtimeKeyVaultName = flowToDeploy.GetTokenString(PortConfigurationSettings.TokenName_RuntimeKeyVaultName);
             Ensure.NotNull(runtimeKeyVaultName, "runtimeKeyVaultName");
 
-            var filePath = ResourcePathUtil.Combine(runtimeConfigBaseFolder, $"{config.Name}-combined.txt");
-            var transformFilePath = await RuntimeStorage.SaveFile(filePath, rulesCode.Code);
             var secretName = $"{config.Name}-transform";
-            var transformFileSecret = await KeyVaultClient.SaveSecretAsync(runtimeKeyVaultName, secretName, transformFilePath, Configuration.TryGet(Constants.ConfigSettingName_SparkType, out string sparkType) ? sparkType : null);
+            Configuration.TryGet(Constants.ConfigSettingName_SparkType, out string sparkType);
+            var uriPrefix = (sparkType != null && sparkType == Constants.SparkTypeDataBricks) ? Constants.PrefixSecretScope : Constants.PrefixKeyVault;
```
vijayupadya: This logic to compute uriPrefix is now in multiple places. Can we consolidate by adding a helper method to KeyVaultClient and call it from everywhere its needed including in KeyVaultClient. Only KeyVaultClient should know the nuances of uri based on sparkType.

kjcho-msft: Makes sense. Created a helper and updated to use it.

PR #116, opened 2019-08-12
Add null checks for guiConfig() as some custom configs don't have a gui section

kjcho-msft:


Services/DataX.Config/DataX.Config.Test/RuntimeConfigGenerationBatchTest.cs
```diff
@@ -88,7 +88,7 @@ public override int GetOrder()
             public override async Task<string> Process(FlowDeploymentSession flowToDeploy)
             {
                 var guiConfig = flowToDeploy.Config.GetGuiConfig();
-                if (guiConfig.Input.Mode == Constants.InputMode_Batching)
+                if (guiConfig?.Input?.Mode == Constants.InputMode_Batching)
```
carlbrochu: anyway to refactor this so future impl don't have to remember to do this?

kjcho-msft: Good point but it's not easy to create a helper method since the way to use the guiConfig varies quite a bit depending on where it's used. But the test I added here should catch this kind of issue.

PR #120, opened 2019-08-15
Codecoverage: filter test binaries and force DataX.Config.dll to build pdbs

carlbrochu:
The Config.dll wouln't be included in the instrumented binaries until I applied a DebugType parameter in the project to force building a pdb.  This doesn't have any other effect.

I also added a filter for the test binaries with a plural filename.


Services/DataX.Config/DataX.Config.LivyClient.Test/JobOperationTest.cs
```diff
@@ -36,6 +36,51 @@ public async Task TestStartJob()
             Assert.AreEqual(expected: JobState.Running, actual: refreshJobInfo.JobState);
         }
 
+        [TestMethod]
```
kjcho-msft: I think I saw these changes in your previous PR?

Services/codecoverage.runsettings
```diff
@@ -0,0 +1,25 @@
...
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="Code Coverage"
+      uri="datacollector://Microsoft/CodeCoverage/2.0" assemblyQualifiedName="Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector,
+ Microsoft.VisualStudio.TraceCollector, Culture=neutral">
+        <Configuration>
+          <CodeCoverage>
+            <ModulePaths>
+              <Include>
+                <ModulePath>.*\DataX.*.dll</ModulePath>
```
kjcho-msft: Should we include Flow.*.exe as well? e.g. Flow.InteractiveQueryService.exe

Services/codecoverage.runsettings
```diff
@@ -0,0 +1,25 @@
...
+              </Include>
+              <Exclude>
+                <ModulePath>.*tests.dll</ModulePath>
+                <ModulePath>.*test.dll</ModulePath>
```
kjcho-msft: Do we need both *tests.dll and *test.dll?

carlbrochu: Yes we have both being built today.

carlbrochu: And didn't want to do that many renames in this pr

PR #127, opened 2019-08-20
Adding JobRunner Service and the first DataX mainline job that calls …

s-tuli:
Adding JobRunner Service and the first DataX mainline job that calls into the ScenarioTester. All sensitive info is in the KeyVault.

### DIFF
--- a/Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/BaseTableEntity.cs
+++ b/Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/BaseTableEntity.cs
@@ -1,0 +1,48 @@
+﻿// *********************************************************************
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License
+// *********************************************************************
+
+using Microsoft.Azure.Cosmos.Table;
+using System;
+using System.Collections.Generic;
+
+namespace DataX.Utilities
+{
+    /// <summary>
+    /// CosmosTable base class
+    /// </summary>
+    public class BaseTableEntity : TableEntity
+    {
+        // azure storage limits
+        private const long _MAX_FIELD_LENGTH = 65536; // 64KB
+        private const long _MAX_TOTAL_FIELD_LENGTH = 1048576; // 1MB
+
+        public override IDictionary<string, EntityProperty> WriteEntity(OperationContext operationContext)
+        {
+
+            var properties = base.WriteEntity(operationContext);
+
+            var totalLength = 0;
+            foreach (var property in properties)
+            {
+                if (property.Value.PropertyType == EdmType.String && property.Value.StringValue != null)
+                {
+                    var value = property.Value.StringValue;
+                    totalLength += value.Length;
+                    if (value.Length > _MAX_FIELD_LENGTH)
+                    {
+                        var e = new ArgumentException("The maximum length of a single field (64KB) has been exceeded", property.Key);
+                        throw e;
+                    }
+                    else if (totalLength > _MAX_TOTAL_FIELD_LENGTH)
+                    {
+                        throw new ArgumentException("The maximum length of all fields (1MB) has been exceeded");
+                    }
+                }
+            }
+
+            return properties;
+        }
+    }
+}
--- a/Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/StorageTableBase.cs
+++ b/Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/StorageTableBase.cs
@@ -1,0 +1,81 @@
+﻿// *********************************************************************
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License
+// *********************************************************************
+
+using DataX.Contract;
+using Microsoft.Azure.Cosmos.Table;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace DataX.Utilities
+{
+    /// <summary>
+    /// Utility base class for CosmosTable
+    /// </summary>
+    /// <typeparam name="TEntity"></typeparam>
+    public abstract class StorageTableBase<TEntity> where TEntity: TableEntity, new()
+    {
+        public StorageTableBase(CloudTable table)
+        {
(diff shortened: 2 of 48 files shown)


Services/JobRunner/ObjectModel/ScheduledJobTable.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License
+// *********************************************************************
+
+using DataX.Utilities;
+using System;
+using System.Composition;
+using System.Threading.Tasks;
+using DataX.Contract;
+
+namespace JobRunner
```
carlbrochu: Should this be DataX.Jobrunner?

Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/StorageTableBase.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using DataX.Contract;
+using Microsoft.Azure.Cosmos.Table;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace DataX.Utilities
+{
+    /// <summary>
+    /// Utility base class for CosmosTable
```
carlbrochu: Is this CosmosDB?

s-tuli: This is CosmosDB Table. The traditional CosmosDB that we use for Data Accelerator has Api=-Core (sql) where as this has api=AzureTable, something we set at the time of creation of the resourceL cosmosDB.

Services/JobRunner/ObjectModel/ScheduledJobTable.cs
s-tuli: Would like to keep it JobRunner since we would want this to be generic at some point in the near future and perhaps create a NuGet such that other teams can use it. Thoughts?

Services/DataX.Utilities.JobRunner/StorageUtils.cs
```diff
@@ -0,0 +1,129 @@
...
+using Microsoft.Azure.Storage.Blob;
+using Microsoft.Azure.Storage.File;
+using System;
+using System.Diagnostics;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace DataX.Utilities
+{
+    public static class StorageUtils
```
vijayupadya: Can you add these to existing storage utils instead of creating a new one?

Services/JobRunner/InstanceExportDescriptorProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Collections.Generic;
+using System.Composition.Hosting.Core;
+
+namespace JobRunner
+{
+    /// <summary>
+    /// Custom export descriptor provider class to add existing instances to MEF container
+    /// </summary>
+    public class InstanceExportDescriptorProvider<TValue> : ExportDescriptorProvider
```
vijayupadya: There is already an existing implementation for this, can you just repurpose the existing one instead of creating a dup class?

Services/JobRunner/ObjectModel/ScheduledJobTable.cs
carlbrochu: ok

Services/JobRunner/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections.Generic;
+using System.Composition.Hosting;
+
+namespace JobRunner
+{
+    public static class ServiceCollectionExtensions
```
vijayupadya: A variant of this class exists in flow.management service. We should reuse code and avoid dup

Services/JobRunner/StartupHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.AspNetCore;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+
+namespace JobRunner
+{
+    /// <summary>
+    /// Helper that gets called from startup for the JobRunner service.
+    /// </summary>
+    public static class StartupHelper
```
vijayupadya: Can we reuse DataXServiceStartup ?

Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/BaseTableEntity.cs
```diff
@@ -0,0 +1,48 @@
+﻿// *********************************************************************
```
kjcho-msft: Isn't this a dup of Services/DataX.Utilities.JobRunner/BaseTableEntity.cs?

Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/StorageTableBase.cs
```diff
@@ -0,0 +1,81 @@
+﻿// *********************************************************************
```
kjcho-msft: this is also a dup of Services/DataX.Utilities.JobRunner/StorageTableBase.cs

s-tuli: Updated

Services/DataX.Utilities.JobRunner/DataX.Utilities.JobRunner/BaseTableEntity.cs
s-tuli: Updated

Services/JobRunner/StartupHelper.cs
s-tuli: DataXServiceStartup is very tightly coupled with DataX code.

Services/DataX.Utilities.JobRunner/StorageUtils.cs
s-tuli: Updated. Added these to the existing project DataX.Utility.Blob

PR #143, opened 2019-09-17
Enable rerunning jobs that were previously in error state

rohit489:


Services/DataX.Config/DataX.Config/InternalService/SparkJobOperation.cs
```diff
@@ -237,7 +237,7 @@ private async Task<SparkJobConfig> EnsureJobState(string jobName, int retries =
                 {
                     //Get Job data
                     var job = await SyncJobState(jobName);
-                    if (job.SyncResult.JobState == JobState.Idle || job.SyncResult.JobState == JobState.Success)
+                    if (job.SyncResult.JobState == JobState.Idle || job.SyncResult.JobState == JobState.Success || job.SyncResult.JobState == JobState.Error)
```
carlbrochu: Is it possible to add a test that covers these cases?

rohit489: added unit test

PR #149, opened 2019-09-25
Remove databricks token from APIs

rohit489:
Remove databricks token from APIs. Create new save button for databricks token. Fix delete kernel API


Services/DataX.Flow/DataX.Flow.InteractiveQuery/InteractiveQueryManager.cs
```diff
@@ -122,8 +122,9 @@ public async Task<ApiResult> GetSampleInputFromQuery(JObject queryObject)
         /// </summary>
         /// <param name="kernels">The list of kernels to be deleted</param>
         /// <param name="subscriptionId">subscriptionId as passed in from the frontend</param>
+        /// <param name="flowName">flow name</param>
         /// <returns>Returns success or error message as the case maybe</returns>
-        public async Task<ApiResult> DeleteKernelList(List<string> kernels)
+        public async Task<ApiResult> DeleteKernelList(List<string> kernels, string flowName)
```
carlbrochu: this still changes api signature right?

kjcho-msft: Can you clarify why flowName is needed here? If I remembered correctly, this API just deletes all kernels specified in the list.

Services/DataX.Flow/DataX.Flow.InteractiveQuery/Databricks/DatabricksKernel.cs
```diff
@@ -58,6 +58,10 @@ public string ExecuteCode(string code)
             var request = new HttpRequestMessage(HttpMethod.Post, url) { Content = new FormUrlEncodedContent(nvc) };
             var response = client.SendAsync(request).Result;
             var responseString = response.Content.ReadAsStringAsync().Result;
+            if (!response.IsSuccessStatusCode)
```
kjcho-msft: comment?

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
```diff
@@ -73,6 +78,35 @@ export default class InfoSettingsContent extends React.Component {
             </div>
         );
     }
+
+    renderSaveButton() {
+        //Will be enabled when user has access 
+        const enableButton = this.props.saveFlowButtonEnabled;
```
kjcho-msft: Do you use the same condition as FlowDefinitionPanel? For this particular one, should we also check if the token field is not empty?

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
```diff
@@ -73,6 +78,35 @@ export default class InfoSettingsContent extends React.Component {
...
+                Save
+            </DefaultButton>
+        );
+    }
+
+    saveFlowAndInitializeKernel() {
+        Q(
+            this.props.onSaveDefinition()
+        )
+            .then(() => {
```
kjcho-msft: indentation?

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
```diff
@@ -73,6 +78,35 @@ export default class InfoSettingsContent extends React.Component {
...
+                this.props.getKernel();
+            })
```
kjcho-msft: As we chatted, let's move this to flowDefinitionPanel.

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowSelectors.js
```diff
@@ -235,11 +235,12 @@ export const getEnableLocalOneBox = createSelector(
...
 export const validateFlowInfo = createSelector(
     getFlowDisplayName,
+    getFlowDatabricksToken,
     validateInfo
 );
 
-function validateInfo(displayName) {
-    return displayName && displayName.trim() !== '';
+function validateInfo(displayName, databricksToken) {
+    return displayName && displayName.trim() !== '' && databricksToken && databricksToken.startsWith('secretscope://');
```
kjcho-msft: How does a new flow experience look like? Should we save it first to generate a secretscope URI and then deploy button will be enabled?

Services/DataX.Flow/DataX.Flow.InteractiveQuery/InteractiveQueryManager.cs
rohit489: Correct. FlowName is required to determine which secret (containing databricks token) to look for in the keyvault

rohit489: FlowName is required to determine which secret (containing databricks token) to look for in the keyvault

Services/DataX.Flow/DataX.Flow.InteractiveQuery/Databricks/DatabricksKernel.cs
rohit489: added

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
rohit489: Added a check to make sure token field is not empty

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
rohit489: fixed

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
rohit489: Moved to flowDefinitionPanel

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowSelectors.js
rohit489: Yes. We would need to first save and only then the deploy button will be enabled

kjcho-msft: Then this changes the user experience for a new flow. Can we just delete this  databricksToken.startsWith('secretscope://' from the conditions?

Services/DataX.Flow/DataX.Flow.InteractiveQuery/InteractiveQueryManager.cs
msandg: will this break existing clients?

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowSelectors.js
kjcho-msft: And you need to check the Spark type since databricksToken is always empty/null for HDInsight here as well.

rohit489: Since user has to always save the token before they can deploy, we need the databricksToken.startsWith('secretscope://' condition. Added check for spark type so that it works for HDInsight scenario

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowSelectors.js
```diff
@@ -235,11 +240,17 @@ export const getEnableLocalOneBox = createSelector(
...
+    getFlowDatabricksToken,
+    getFlowIsDatabricksSparkType,
     validateInfo
 );
 
-function validateInfo(displayName) {
-    return displayName && displayName.trim() !== '';
+function validateInfo(displayName, databricksToken, isDatabricksSparkType) {
+    if (isDatabricksSparkType) {
+        return displayName && displayName.trim() !== '' && databricksToken && databricksToken.startsWith('secretscope://');
```
kjcho-msft: Let's create/use a constant for "secretscope://"

Services/DataX.Flow/DataX.Flow.InteractiveQuery/InteractiveQueryManager.cs
rohit489: If users update the service then they would have to update the website as well

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
```diff
@@ -73,6 +77,26 @@ export default class InfoSettingsContent extends React.Component {
             </div>
         );
     }
+
+    renderSaveButton() {
+        //Will be enabled when user has access
```
kjcho-msft: //Will be enabled when user has access and the token is not empty

Website/Packages/datax-pipeline/src/modules/flowDefinition/flowSelectors.js
rohit489: Done

Services/DataX.Flow/DataX.Flow.InteractiveQuery/Databricks/DatabricksKernel.cs
```diff
@@ -58,6 +58,11 @@ public string ExecuteCode(string code)
...
+            {
+                //If request fails then return error string
+                return responseString;
```
vijayupadya: Returning a raw response may not be appropriate as it may contain uncomprehending xml. Instread return only the error code and the message.

Website/Packages/datax-pipeline/src/modules/flowDefinition/components/info/infoSettingsContent.jsx
rohit489: Updated comment

Services/DataX.Flow/DataX.Flow.InteractiveQuery/Databricks/DatabricksKernel.cs
rohit489: In this context the response message displays the actual error message, for eg.: '{"error":"ContextNotFound: "}' . While the error code only says 500 and response phrase says 'Internal server error' without giving detail about the error. Thus returning the response

PR #152, opened 2019-10-03
For databricks by default disable autoscale

rohit489:


Website/Packages/datax-pipeline/src/modules/flowDefinition/flowReducer.js
```diff
@@ -20,8 +20,8 @@ const INITIAL_FLOW_STATE = {
     scale: {
         jobNumExecutors: '4',
         jobExecutorMemory: '1000',
-        jobDatabricksAutoScale: true,
-        jobDatabricksMinWorkers: '1',
+        jobDatabricksAutoScale: false,
+        jobDatabricksMinWorkers: '3',
```
kjcho-msft: should this be 3 ?

rohit489: 3 is the new default value. Users can still move the slider the on the UI to choose a different value

PR #159, opened 2019-10-14
Added unit tests for components in host and utility

vijayupadya:


DataProcessing/datax-host/pom.xml
```diff
@@ -107,7 +107,7 @@ SOFTWARE
         <dependency>
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_2.11</artifactId>
-            <version>2.2.6</version>
+            <version>3.0.8</version>
```
carlbrochu: do we need to rev the versions of all componets to 1.3?

DataProcessing/datax-host/src/main/scala/datax/sink/OutputManager.scala
```diff
@@ -72,6 +72,8 @@ object OutputManager {
     val onInitHandlers = sinkOperators.map(_._5).filter(_!=null)
     var shouldGeneratorProcessedSchema = processedSchemaPath!=null && !processedSchemaPath.isEmpty
 
+
```
carlbrochu: empty?

DataProcessing/datax-host/pom.xml
carlbrochu: @rohit489

rohit489: We should upgrade the version of all the packages to 1.3.0-SNAPSHOT

carlbrochu:
Will reopen when we get back to this

PR #160, opened 2019-10-17
Add a batch sample

kjcho-msft:
Add a batch sample based on the EventHub streaming sample.


DeploymentCloud/Deployment.DataX/Samples/flows/eventhubbatch-product.json
```diff
@@ -0,0 +1,574 @@
...
+			"sparkJobDatabricksAutoScale" : "${guiSparkJobDatabricksAutoScale}",
+			"processedSchemaPath" : null
+		},
+		"jobs" : [
+			{
+				"partitionJobNumber" : "1"
+			}
+		]
+	},
+	"metrics" : {
```
carlbrochu: we still generate those for batch?

DeploymentCloud/Deployment.DataX/Samples/flows/eventhubbatch-product.json
```diff
@@ -0,0 +1,574 @@
...
+			],
+			"jobconfig" : {
+				"jobNumExecutors" : "4",
+				"jobExecutorMemory" : "1000",
+				"jobDatabricksAutoScale" : false,
```
carlbrochu: will this work for hdi and db?

kjcho-msft: yes it should work both.

DeploymentCloud/Deployment.DataX/Samples/flows/eventhubbatch-product.json
kjcho-msft: yes we do. It still send the metrics to eventhub and store them into the cache. We can display them later if needed.

DeploymentCloud/Deployment.DataX/deploySample.ps1
```diff
@@ -179,12 +199,15 @@ function Add-CosmosDBData ([string]$filePath) {
...
 
-    $tokens.Add('subscriptionId', $subscriptionId )
+    $tokens.Add('subscriptionId', $subscriptionId)
     $tokens.Add('resourceLocation', $resourceGroupLocation)
     $tokens.Add('iotHubName', $iotHubName)
     $tokens.Add('kafkaName', $kafkaName)
     $tokens.Add('kafkaEventHubNamespaceName', $kafkaEventHubNamespaceName)
     $tokens.Add('clientSecretPrefix', $clientSecretPrefix)
+
+    $deploymentDate = Get-Date -Format "yyyy-MM-dd"
```
rohit489: When running the ARM template if the date changes, will the sample still work? eg. ARM template is started at 11 PM and it completes at 2 AM the next day, will the sample still work ?

kjcho-msft: yes it will still work. This datetime will be cached only when the deploysample.ps1 executes.

PR #161, opened 2019-10-18
Scenario Tester Deployment Scripts

jozavala-msft:
This PR adds an azure deployment json file to be deployed in a target environment. The basic parameters for the deployment scripts is the target subscription, tenant, resource group and product name including other parameters defined in the parameter file. Part of the setup involves creating pre-defined values for scenario tester into the storage account for the first test api call (ie. create the referenced data, upload udf to the blob and create a payload with those references). It receives as input the application id and secret scenario tester will act on behalf of to do the api calls.

### DIFF
--- a/DeploymentCloud/Deployment.JobRunner/deployResources.ps1
+++ b/DeploymentCloud/Deployment.JobRunner/deployResources.ps1
@@ -1,0 +1,86 @@
+param(
+    [string]
+    $tenantId,
+    [string]
+    $subscriptionId,
+    [string]
+    $resourceGroupName,
+    [string]
+    $applicationId,
+    [string]
+    $appSecretKey,
+    [string]
+    $productName,
+    [string]
+    $sparkType,
+    [string]
+    $sparkStorageAccountName,
+    [string]
+    $kvSparkName,
+    [string]
+    $kvServicesName,
+    [string]
+    $flowName='scenariotest'
+)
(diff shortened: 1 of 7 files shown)


DeploymentCloud/Deployment.JobRunner/deployResources.ps1
```diff
@@ -0,0 +1,89 @@
...
+    [string]
+    $tenantId,
+    [string]
+    $subscriptionId,
+    [string]
+    $resourceGroupName,
+    [string]
+    $applicationId,
+    [string]
+    $appSecretKey,
```
rohit489, line 11: how do we get the values of these parameters

rohit489:
Can you add a readme that instructs how to deploy scenario tester

DeploymentCloud/Deployment.JobRunner/deployResources.ps1
jozavala-msft: Those values are retrieved from previous data-accelerator deployment. More specifically, those are the chosen values from common.parameters.txt and admin.parameters.txt values, and then ones generated for the deployment script such the keyvault prefix. At this moment, the user should look for those values at the target environment where data-accelerator is deployed. I added those details at the README.MD file

jozavala-msft:
@rohit489 I added the readme file, can you please take a look again?

PR #166, opened 2020-01-25
Added check for mdbc PSModule version

ankisho:
**Issue**: Deployment breaks due to New-MdbcQuery cmdlet doesn't found.

**Fix**:
Mdbc PSModule doesn't support New-MdbcQuery cmdlet from v6.0.0. we have to make sure correct Mdbc module (v5.1.4) is installed in the system during deployment.
Deployment uses New-MdbcQuery cmdlet to setup CosmosDb. 
$qry = New-MdbcQuery -Name "_id" -Exists

### DIFF
--- a/DeploymentCloud/Deployment.Common/deployResources.ps1
+++ b/DeploymentCloud/Deployment.Common/deployResources.ps1
@@ -133,6 +133,15 @@ function Install-Modules {
                 $moduleInstalled = $true
                 Install-Module -Name $_ -Force -AllowClobber -Scope CurrentUser -Repository PSGallery
             }
+	    
+	    # Since, Mdbc PS module doesn't support New-MdbcQuery cmdlet from v6.0.0, we have to make sure correct Mdbc module is installed.
+            # https://github.com/nightroman/Mdbc/blob/master/Release-Notes.md#v600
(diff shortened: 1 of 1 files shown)


DeploymentCloud/Deployment.Common/deployResources.ps1
```diff
@@ -133,6 +133,15 @@ function Install-Modules {
                 $moduleInstalled = $true
                 Install-Module -Name $_ -Force -AllowClobber -Scope CurrentUser -Repository PSGallery
             }
+	    
+	    # Since, Mdbc PS module doesn't support New-MdbcQuery cmdlet from v6.0.0, we have to make sure correct Mdbc module is installed.
+            # https://github.com/nightroman/Mdbc/blob/master/Release-Notes.md#v600
+            if ($_ -eq 'mdbc' -and !(Get-InstalledModule -Name $_ -RequiredVersion $modules.Item($_) -ErrorAction SilentlyContinue)) {
```
rohit489: the if condition above will try to install mdbc module. Lets add a check to skip the previous if condition for mdbc package

ankisho: New approach: Install the required version of all 3 PS modules and import them.

PR #172, opened 2020-08-03
add restartall API to the flow service

msandg:
all restartall API


jozavala-msft:
I like this implementation. I want to confirm with @rohit489 , is this new endpoint a test candidate for this suite of tests => https://github.com/microsoft/data-accelerator/tree/master/Tests/DataXScenarios/DataXScenarios ? If thats the case, i think it would be good to add a test call to this new endpoint there

msandg:
let's use another PR https://github.com/microsoft/data-accelerator/pull/173

PR #173, opened 2020-08-04
add restartall API to the flow service

msandg:


msandg:
Jose's comment:

I like this implementation. I want to confirm with @rohit489 , is this new endpoint a test candidate for this suite of tests => https://github.com/microsoft/data-accelerator/tree/master/Tests/DataXScenarios/DataXScenarios ? If thats the case, i think it would be good to add a test call to this new endpoint there

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
```diff
@@ -422,6 +422,25 @@ public async Task<ApiResult> RestartJob([FromBody] string jobName)
             }
         }
 
+        [HttpGet]
```
rohit489, line 425: Should we make it HttpPost. Will also help if take take an input list of jobnames to restart

Services/DataX.Config/DataX.Config/PublicService/JobOperation.cs
```diff
@@ -49,6 +49,19 @@ public async Task<Result> RestartJob(string jobName)
             return await this.SparkJobOperation.RestartJobWithRetries(jobName, timeout, _DefaultRetryInterval);
         }
 
+        public async Task<Result[]> RestartAllJobs()
+        {
+            var jobs = await SparkJobData.GetAll();
+            var results = new List<Result>();
+            foreach (var job in jobs)
+            {
+                var result = await RestartJob(job.Name);
```
rohit489, line 58: We should add a check to verify if the job was restarted fine. If we restart a lot of jobs in a small time interval we have seen failures in job restart where HDInsight fails to allocate resources to the jobs

rohit489: We should also add a retry logic in case of failures

rohit489: As we restart a batch of jobs in a given list of jobs, we can call SyncJobStateByNames() to check the status of each job in the batch and once all the jobs are in running state then proceed to next batch in the list. If some job was not in running state after a threshoold time period then we can issue a restart again for that job

rohit489:
> Jose's comment:
> 
> I like this implementation. I want to confirm with @rohit489 , is this new endpoint a test candidate for this suite of tests => https://github.com/microsoft/data-accelerator/tree/master/Tests/DataXScenarios/DataXScenarios ? If thats the case, i think it would be good to add a test call to this new endpoint there

Yes makes sense to add this to scenario tester as well like @jozavala-msft 

> Jose's comment:
> 
> I like this implementation. I want to confirm with @rohit489 , is this new endpoint a test candidate for this suite of tests => https://github.com/microsoft/data-accelerator/tree/master/Tests/DataXScenarios/DataXScenarios ? If thats the case, i think it would be good to add a test call to this new endpoint there

Yes makes sense to add it to scenario tester like @jozavala-msft  pointed

Services/DataX.Config/DataX.Config/PublicService/JobOperation.cs
msandg: 1. added a new API to wait for jobs up
2. added retry logics in the wait API

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
```diff
@@ -463,5 +481,42 @@ public async Task<ApiResult> SyncJobsByNames([FromBody] string[] jobNames)
...
+
+                int retryTimes = 0;
+                SparkJobFrontEnd[] jobOpResult = null;
+                while (retryTimes++ < maxRetryTimes)
+                {
+                    jobOpResult = jobNames != null && jobNames.Any() ?
+                                await _jobOperation.SyncJobStateByNames(jobNames)
+                                : await _jobOperation.SyncAllJobState();
+
+                    var startingJobs = jobOpResult.Where(r => r.JobState == JobState.Starting);
```
rohit489, line 504: In case resources donot get allocated to the spark job on HDInsight, the state changes to idle. In this case job submission needs to be retired too

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
```diff
@@ -463,5 +481,42 @@ public async Task<ApiResult> SyncJobsByNames([FromBody] string[] jobNames)
...
+                    if (!startingJobs.Any())
```
rohit489, line 505: in case the job is not running, shouldn't we issue a restart again for that job?

jozavala-msft (approved):
We agreed that manual testing is enough for this feature. We are covering automated testing in another PR, and we don't need to add scenario tester code here. LGTM. Lets wait for @rohit489 approval as well.

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
guiyang2011: yes, it would restart when job not running

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
msandg: have updated the logic to restart the job state is idle, starting or failed

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
msandg: yes, will restart it

Services/DataX.Flow/Flow.ManagementService/Controllers/FlowManagementController.cs
```diff
@@ -422,6 +422,49 @@ public async Task<ApiResult> RestartJob([FromBody] string jobName)
             }
         }
 
+        [HttpPost]
+        [Route("job/restartallwithretries")]
+        [DataXWriter]
+        public async Task<ApiResult> RestatAllWithRetries([FromBody] RestatAllWithRetriesRequest request)
```
rohit489, line 428: Instead of taking parameter of type RestatAllWithRetriesRequest, lets use JObject. See "public async Task<ApiResult> SaveFlow([FromBody]JObject config)" for reference

Services/DataX.Flow/Flow.ManagementService/Controllers/RestartAllWithRetriesRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Flow.Management.Controllers
```
rohit489, line 1: Instead of adding model class in controller, add it in DataX.Config.ConfigDataModel

rohit489 (approved):

PR #176, opened 2020-09-19
Add keyboard navigation 

rohit489:
Add keyboard navigation to tabs on flow details page  and start/stop/restart button on jobs page


Website/Packages/datax-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
```
carlbrochu, line 1: looks straightforward enough. can @barrytu or @dennygu review?  do we need to rev the version\n

carlbrochu, line 1: looks straightforward enough. can @barrytu or @dennygu review?  do we need to rev the version\n

jozavala-msft (approved):
Looks good to me.

PR #182, opened 2021-02-12
Fix metrics header

rohit489:


carlbrochu:
any other "header" that should be a div?

carlbrochu (approved):

rohit489:
> any other "header" that should be a div?

With this fix, there should not be any more div that need to be converted to heading

PR #188, opened 2021-09-24
add abfss protocol for adls gen2 experiment

birgithi:


DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
```diff
@@ -66,7 +66,7 @@ object HadoopClient {
   }
 
   /***
-    * get the name of storage account from a wasb-format path
+    * get the name of storage account from a wasb-format or abfss-format path
     * @param path a hdfs path
     * @return the storage account name if there is storage account name in the wasbs/wasb path, else null
     */
```
birgithi, line 72: Can we rename getWasbStorageAccount to something more generic since this no longer is only for wasb format? I wasn't sure if it was okay to rename this or if that will break something.

DataProcessing/datax-host/src/main/scala/datax/input/BlobPointerInput.scala
```diff
@@ -39,7 +39,7 @@ object BlobPointerInput {
     (new StructType).add("BlobPath", StringType)
   }
 
-  private val saRegex = s"""wasbs?://[\w-]+@([\w\d]+)\${BlobProperties.BlobHostPath}/.*""".r
+  private val saRegex = s"""(wasbs?|abfs?)://[a-zA-Z0-9-_]+@([a-zA-Z0-9-_]+)(${BlobProperties.BlobHostPath}||${BlobProperties.DfsHostPath})/.*""".r
```
birgithi, line 42: I had to change the regex here since I was getting an error using \w. I want to be sure that this doesn't break any existing functionality.

DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
carlbrochu: should be ok as long as you rename all instances

DataProcessing/datax-host/src/main/scala/datax/input/BlobPointerInput.scala
jozavala-msft: I was working in a change on that exact line in another PR, however i prefer your fix over mine. I think it would be a good idea to add a unit test just to ensure \w error is gone. Can you add a very basic BlobPointerInput test to make sure the class is initialized without exceptions, take as example the unit test added at https://github.com/microsoft/data-accelerator/pull/192/files , then i can remove the changes on this file/unit test from my PR

DataProcessing/datax-host/pom.xml
```diff
@@ -118,13 +118,18 @@ SOFTWARE
...
+            <version>1.2.3-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>com.microsoft.datax</groupId>
             <artifactId>datax-utility_2.4_2.11</artifactId>
             <version>1.2.2-SNAPSHOT</version>
         </dependency>
+		<dependency>
+		  <groupId>com.microsoft.azure</groupId>
+		  <artifactId>azure-keyvault-cryptography</artifactId>
```
carlbrochu, line 130: @rohit489  is this the right component?

DataProcessing/datax-host/pom.xml
```diff
@@ -118,13 +118,13 @@ SOFTWARE
         <dependency>
             <groupId>com.microsoft.datax</groupId>
             <artifactId>datax-core_2.4_2.11</artifactId>
-            <version>1.2.3-SNAPSHOT</version>
+            <version>1.2.4-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>com.microsoft.datax</groupId>
             <artifactId>datax-utility_2.4_2.11</artifactId>
             <version>1.2.3-SNAPSHOT</version>
```
rohit489, line 126: The version of all the datax dependency packages would change to 1.2.4-SNAPSHOT

DataProcessing/datax-host/pom.xml
```diff
@@ -118,13 +118,13 @@ SOFTWARE
         <dependency>
             <groupId>com.microsoft.datax</groupId>
             <artifactId>datax-core_2.4_2.11</artifactId>
-            <version>1.2.3-SNAPSHOT</version>
+            <version>1.2.4-SNAPSHOT</version>
```
rohit489, line 121: Please rev the version of all the datax packages to "1.2.4-SNAPSHOT" including datax-keyvault and datax-udf-samples. We normally keep the same versions for all the packages

DataProcessing/datax-host/pom.xml
birgithi: Fixed

DataProcessing/datax-host/pom.xml
birgithi: I updated all of the versions to 1.2.4-SNAPSHOT

rohit489 (approved):

PR #190, opened 2021-09-28
explicitly inlude azure-keyvault-cryptography package

rohit489:
"azure-keyvault-cryptography" package which is a system dependency of "azure-keyvault" package no longer gets automatically intalled. hence explicitly including it pom

### DIFF
--- a/DataProcessing/datax-host/pom.xml
+++ b/DataProcessing/datax-host/pom.xml
@@ -126,7 +126,7 @@ SOFTWARE
             <version>1.2.3-SNAPSHOT</version>
         </dependency>
         <dependency>
-            <groupId>com.microsoft.datax</groupId>
+            <groupId>com.microsoft.azure</groupId>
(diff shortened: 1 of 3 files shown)


DataProcessing/datax-host/pom.xml
```diff
@@ -125,6 +125,11 @@ SOFTWARE
             <artifactId>datax-utility_2.4_2.11</artifactId>
             <version>1.2.3-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.datax</groupId>
+            <artifactId>azure-keyvault-cryptography</artifactId>
```
carlbrochu, line 130: Is this needed if below exists

carlbrochu (approved):

DataProcessing/datax-host/pom.xml
rohit489: yes this is needed in each respective package to be able to build each package separately

PR #191, opened 2021-09-29
Avoid building the regex if it is not actually used

jozavala-msft:
Under some circumstances, if the regex is started it crashes with  scala.StringContext$InvalidEscapeException. This can be avoided if the regex is provided through configuration using sourceidregex property. Make regex variable lazy, so it is instantiated when it needs to be used only.


jozavala-msft:
Included in https://github.com/microsoft/data-accelerator/pull/192

PR #223, opened 2022-04-15
Upgrade to Kafka components version 2.8.1

carlbrochu:
Upgrading to newer Kafka components.

Downgrading jackson component for now to fix the build


sumabh-msft (approved):
looks good to me

PR #228, opened 2022-05-06
drop blob without renaming in HadoopClient

yinyuwang:
in order to solve the issue that no events triggered when blob file is created using abfs protocol, we should cancel the renaming step and drop the file directly to the location

### DIFF
--- a/DataProcessing/datax-host/src/main/scala/datax/checkpoint/EventhubCheckpointer.scala
+++ b/DataProcessing/datax-host/src/main/scala/datax/checkpoint/EventhubCheckpointer.scala
@@ -57,7 +57,7 @@ object EventhubCheckpointer {
     }
 
     HadoopClient.writeHdfsFile(checkpointFile,
-      offsets.map(v=>v._1+OffsetTokenSeparator+v._2+OffsetTokenSeparator+v._3+OffsetTokenSeparator+v._4+OffsetTokenSeparator+v._5).mkString("\n"), true, false)
+      offsets.map(v=>v._1+OffsetTokenSeparator+v._2+OffsetTokenSeparator+v._3+OffsetTokenSeparator+v._4+OffsetTokenSeparator+v._5).mkString("\n"), overwriteIfExists=true, directWrite=false)
   }
 
   def applyCheckpointsIfExists(ehConf: EventHubsConf, checkpointDir: String) = {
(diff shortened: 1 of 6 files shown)


DataProcessing/datax-host/src/main/scala/datax/checkpoint/EventhubCheckpointer.scala
```diff
@@ -57,7 +57,7 @@ object EventhubCheckpointer {
     }
 
     HadoopClient.writeHdfsFile(checkpointFile,
-      offsets.map(v=>v._1+OffsetTokenSeparator+v._2+OffsetTokenSeparator+v._3+OffsetTokenSeparator+v._4+OffsetTokenSeparator+v._5).mkString("\n"), true)
```
carlbrochu, line 60: is behavior the same as before here?

carlbrochu: i think it is, if you can confirm.

DataProcessing/datax-host/src/main/scala/datax/sink/OutputManager.scala
```diff
@@ -85,7 +85,7 @@ object OutputManager {
           val spark = df.sparkSession
           spark.synchronized{
             if(shouldGeneratorProcessedSchema){
-              HadoopClient.writeHdfsFile(processedSchemaPath, new StructType(outputColumns).prettyJson, true)
+              HadoopClient.writeHdfsFile(processedSchemaPath, new StructType(outputColumns).prettyJson, true, false)
```
bryanchen-d, line 88: Let's use named parameter when passing the values here so it will be readable for viewers. so this should be like
```scala
HadoopClient.writeHdfsFile(processedSchemaPath, 
                                          new StructType(outputColumns).prettyJson, 
                                          overwriteIfExists = true, 
                                          directWrite = false)
```

bryanchen-d (approved):

DataProcessing/datax-host/src/main/scala/datax/checkpoint/EventhubCheckpointer.scala
yinyuwang: yes, adding directWrite=false will remain the same as before

PR #229, opened 2022-05-27
fix json4s jackson runtime dependencies error

yinyuwang:
This is to fix the runtime dependency error when fetching access token in KV client. In Spark 2.4, the compatible version of json4s jackson is 3.5.3. The program will run into java.NoSuchMethod error when 4.0.4 is set.

### DIFF
--- a/DataProcessing/datax-keyvault/pom.xml
+++ b/DataProcessing/datax-keyvault/pom.xml
@@ -67,7 +67,7 @@ SOFTWARE
     <dependency>
       <groupId>org.json4s</groupId>
       <artifactId>json4s-jackson_2.11</artifactId>
-      <version>4.0.4</version>
+      <version>3.5.3</version>
     </dependency>
     <!--Azure dependencies-->
     <dependency>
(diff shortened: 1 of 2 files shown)


DataProcessing/datax-keyvault/pom.xml
```diff
@@ -67,7 +67,7 @@ SOFTWARE
     <dependency>
       <groupId>org.json4s</groupId>
       <artifactId>json4s-jackson_2.11</artifactId>
-      <version>4.0.4</version>
+      <version>3.5.3</version>
```
bboonmicrosoft, line 70: What will this do to nova prod?

yinyuwang: Suman will help me test in nova prod

jozavala-msft: I was able to reproduce the error in dddatax backup cluster using the latest bits of master branch. After downgrading json4s the error is gone and the job is not crashing anymore. I suspect some recent change in version in another artifact re-introduced this error after 4.0.4 upgrade. We are completing the testing and i am approving this PR after that

jozavala-msft (approved):

PR #235, opened 2022-06-01
Update dockerfile to pull in dotnet image from mcr

rohit489:
Update dockerfile to pull in dotnet image from mcr.microsoft.com


carlbrochu:
there is one more file under the kubernetese deployment.

rohit489:
> there is one more file under the kubernetese deployment.

Thanks for catching this. Updated the file in kubernetese folder

carlbrochu (approved):

PR #254, opened 2022-06-27
support eventhub rbac

yinyuwang:
This is to support the AAD RBAC-based method of connecting to eventhub. We have an Azure compliance requirement to move off keys / connection strings where possible and instead use AAD RBAC-based authentication


DataProcessing/datax-host/src/main/scala/datax/client/eventhub/EventHubBase.scala
```diff
@@ -21,7 +24,43 @@ object EventHubBase {
...
+    })
+    extracted
+  }
+
+  def buildClientFromConnString(extracted: Map[String, String]): Boolean = {
+    extracted.getOrElse("SharedAccessKeyName", "") != "" &&
+      extracted.getOrElse("SharedAccessKey", "") != ""
+  }
+
   def getOutputClient(connString: String) = {
```
rohit489, line 43: With RBAC enabled for eventhub, a connection string is not a requirement. How would you handle the case when connString is not passed in

rohit489 (approved):

PR #256, opened 2022-06-29
Pointing to each pom and package.json

carlbrochu:
Need to point to exact paths w each pom and package.json

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,11 +6,31 @@
 version: 2
 updates:
   - package-ecosystem: "npm" # See documentation for possible values
-    directory: "/Website" # Location of package manifests
+    directory: "/Website/Website" # Location of package.json
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,11 +6,31 @@
 version: 2
 updates:
   - package-ecosystem: "npm" # See documentation for possible values
-    directory: "/Website" # Location of package manifests
+    directory: "/Website/Website" # Location of package.json
```
rohit489, line 9: Any reason why the other NPM packages are excluded from the list? Following are the missing packages
1. /Website/Packages/datax-jobs
2. /Website/Packages/datax-metrics
3. /Website/Packages/datax-pipeline
4. /Website/Packages/datax-query

carlbrochu: added

rohit489 (approved):

PR #495, opened 2022-08-16
Jozavala/log4j applicationinsight appender

jozavala-msft:
Use AppInsight Log4J appender to send logs to application insights instance.

### DIFF
--- a/DataProcessing/datax-host/pom.xml
+++ b/DataProcessing/datax-host/pom.xml
@@ -51,7 +51,7 @@ SOFTWARE
 
     <groupId>com.microsoft.datax</groupId>
     <artifactId>datax-host_2.4_2.11</artifactId>
-    <version>1.2.12-SNAPSHOT</version>
+    <version>1.2.13-SNAPSHOT</version>
 
     <properties>
         <spark.version>2.4.5</spark.version>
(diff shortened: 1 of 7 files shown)


DataProcessing/datax-host/pom.xml
```diff
@@ -51,7 +51,7 @@ SOFTWARE
 
     <groupId>com.microsoft.datax</groupId>
     <artifactId>datax-host_2.4_2.11</artifactId>
-    <version>1.2.12-SNAPSHOT</version>
+    <version>1.2.13-SNAPSHOT</version>
```
rohit489, line 54: missing version rev for other dependency datax package. like datax-core

DataProcessing/datax-host/src/main/scala/datax/telemetry/AppInsightLogger.scala
```diff
@@ -37,6 +41,11 @@ object AppInsightLogger extends TelemetryService {
       logger.warn("AI Key is not found, AppInsight Sender is OFF")
   }
 
+  private val bAppenderEnabled = (for {
```
rohit489, line 44: what does b stand for in bAppenderEnabled. Please name the val so that is more descriptive.

DataProcessing/datax-host/pom.xml
rohit489: just checked. looks like the {version} is being used for the dependencies so its good

rohit489 (approved):

PR #532, opened 2022-09-08
convert to 15 mins batch

yinyuwang:


DataProcessing/datax-host/src/main/scala/datax/fs/HadoopClient.scala
```diff
@@ -521,6 +520,23 @@ object HadoopClient {
       Iterator.empty
   }
 
+  /**
+   * list file objects under a folder
+   *
+   * @param folder path to the specified folder
+   * @return a list of file paths under the folder
+   */
+  def listFileObjects(folder: String): Iterator[FileStatus] = {
```
rohit489, line 529: Refactor the code. Duplicate code from listFiles()

DataProcessing/datax-host/src/main/scala/datax/host/BlobBatchingHost.scala
```diff
@@ -124,22 +129,46 @@ object BlobBatchingHost {
...
 
   // write a tracker file in specific folder with format _SUCCESS_yyyy_MM_dd_HH
-  def writeTracker(trackerFolder: String, dt: Timestamp): Unit = {
+  def writeTracker(trackerFolder: String, dt: Timestamp, dtEnd: Timestamp): Unit = {
     // first create folder if not exists
     HadoopClient.createFolder(trackerFolder)
-    val dateFormat = new SimpleDateFormat("yyyy_MM_dd_HH")
+    var fmt = "yyyy_MM_dd_HH"
+    // if the interval is not 1h, the use the format yyyy_MM_dd_HH_mm
+    if (dtEnd.getTime / 1000 - dt.getTime / 1000 + 1 != 3600) {
```
rohit489, line 145: Why have 2 different formats? In case the interval is 1 hour, cant we still have the format as yyyy_MM_dd_HH_mm where mm value will be 00 ?

yinyuwang: This is to keep the logic on both sides consistent. https://msazure.visualstudio.com/One/_git/DnA-M365A-DataPipeline/pullrequest/6722026?_a=files&path=/src/JS/EdpConfigManager/src/partners/ddvcore/generators/cosmosEgressParamsGenerator.ts

rohit489 (approved):

PR #1290, opened 2024-03-06
migrating pipelines to 1ES yaml

v-jeevanj:
Migrating classic pipelines to 1ES yaml.


v-jeevanj:
Abandoning this as we moved to using GitHub workflow for the OSS repo.

PR #1460, opened 2024-08-28
Migrate MSI Server to MSAL

Ed-Maeng:
#### PR Summary
- Migrate MSI Sever to MSAL to help better fetch access token for CP jobs
- ADAL is no longer supported
- MSAL should be more up to date with token caching and refreshes
-  `AcquireTokenForClient` can rely on MSAL's internal caching: https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/acquire-token-silently

#### Related PR
- DataX Change: https://msazure.visualstudio.com/One/_git/DevDiv-EDP-DataPipeline/pullrequest/10763200
- EDP Change: https://msazure.visualstudio.com/One/_git/DnA-M365A-DataPipeline/pullrequest/10763902


DeploymentCloud/Deployment.Common/scripts/startmsiserverservice.sh
```diff
@@ -1,6 +1,35 @@
...
 sudo systemctl daemon-reload
+
+echo "Enable the service to start on boot"
 sudo systemctl enable msiserverapp.service
-sudo systemctl start msiserverapp.service
\ No newline at end of file
+sudo systemctl enable msalmsiserverapp.service
+
+echo "Start the service"
+sudo systemctl start msiserverapp.service
```
rohit489, line 32: Why are we starting msiserverapp.service if we are migrating off of it ?

Ed-Maeng: I was planning to keep the old server in a different port for now, just in case we might need to revert the changes. Planning to delete the old server and remove it from the script once we have validated everything works fine in prod.

rohit489: did you confirm if both msal and adal run together ?

Ed-Maeng: Yes! We as long as they have a separate port and service file to run. Dev VS01 cluster is currently running both servers.

rohit489 (approved):

